PR #3, opened 2022-05-09
Perl to 2.26: don't push to scalar

MatBarba:
A simple fix for a deprecated syntax


JAlvarezJarreta (commented):
I'm surprised to see the syntax `\@`, I though the Perl way for dereference a variable was `@$` (or `%$`). That being said, should we change `\%` as well since you are already touching the line?
On another note, can you fix the typo in the commit (and PR) from `2.26` to `5.26`? Just so we don't get newcomers confused if they check this in the future :see_no_evil:

MatBarba:
In this case it's likely an error in the code, since no reference was needed for the array.

I'll have to resubmit a branch to change the commit summary, since commits can't be modified

MatBarba:
Using \% to give a reference to a hash is proper though, so no need to change it

PR #7, opened 2022-06-07
The initial commit to add Sphinx setup and slight reconfiguration of its build scripts

ens-LCampbell:


scripts/setup/docs/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
vsitnik, line 1: do we need make.bat at all?

vsitnik (approved):
I'm ok with this. feel free to merge

scripts/setup/docs/make.bat
vsitnik: (perhaps, we need to fix paths, similar to Makefile)

ens-LCampbell: > do we need make.bat at all?

I don't know as I simply used files generated as specified in Sphinx online setup docs 'sphinx-quickstart'. Perhaps not but I haven't tested this yet but I will test now

PR #15, opened 2022-07-08
Addition of IGNORE rules for cases of pseudogenes linked to 3_prime |…

ens-LCampbell:
… 5_prime UTRs

Updates to valid_structures.conf to account for cases encountered in a MAKER2 annotation in which pseudogenes were defined with CDS, exon and UTRs. 

E.g:
[pseudo_with_UTRs.txt](https://github.com/Ensembl/ensembl-genomio/files/9072562/pseudo_with_UTRs.txt)

### DIFF
--- a/scripts/gff_metaparser/conf/valid_structures.conf
+++ b/scripts/gff_metaparser/conf/valid_structures.conf
@@ -94,9 +94,6 @@ gene/pseudogene/CDS	SUB:load_pseudogene_with_CDS:IGNORE	pseudogene/mRNA:biotype=
 
 pseudogene/pseudogenic_trna/exon	VALID	# will be substituted to pseudogene/tRNA_pseudogene/exon by LoadGFF3
 
-pseudogene/three_prime_UTR      IGNORE:ignore_pseudogene_utr:UNSEEN
-pseudogene/five_prime_UTR       IGNORE:ignore_pseudogene_utr:UNSEEN
-
 # MT related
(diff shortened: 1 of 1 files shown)


scripts/gff_metaparser/conf/valid_structures.conf
```diff
@@ -150,6 +149,9 @@ gene/pseudogene/intron	IGNORE
 gene/trna/intron	IGNORE
 gene/sequence_variant	IGNORE
 
+## IGNORED STRANGE OR INVALID MODELS / PARTS
+pseudogene/three_prime_UTR	IGNORE:ignore_pseudogene_utr:UNSEEN
```
vsitnik, line 153: could you make them just [always] "IGNORE", under no flags ?

vsitnik: or do you prefer to have a flag to report them by default?

ens-LCampbell: If we set them to IGNORE outright, they are not going to be reported in the gff_stats.err log file right ? I think let's leave it for now. Its not something seen very often and reporting it then setting the flag in conf isn't difficult to do.

vsitnik: makes sense. thanks!

vsitnik (approved):
ok

PR #24, opened 2022-10-25
Optimize slice cache

MatBarba:
Retrieving translation sequences can be very slow when the seq_regions are very long, because then we can't take advantage of the cache.

These commits change the loops from all transcripts, to all slices, all of them split up so that they be cached.


Dishalodha (approved):
Seems good to me 👍

PR #32, opened 2022-12-04
Chunks

vsitnik:
Enabling chunking of input dna fasta when "sequence_data_chunck"  is enabled (>= 50_000)

### DIFF
--- a/lib/python/ensembl/brc4/runnable/load_sequence_data.py
+++ b/lib/python/ensembl/brc4/runnable/load_sequence_data.py
@@ -45,7 +45,7 @@ class load_sequence_data(eHive.BaseRunnable):
         """
         return {
             # relative order of the coord_systems types to infer their rank from
-            'cs_order' : 'chunk,contig,supercontig,non_ref_scaffold,scaffold,primary_assembly,superscaffold,linkage_group,chromosome',
+            'cs_order' : 'ensembl_internal,chunk,contig,supercontig,non_ref_scaffold,scaffold,primary_assembly,superscaffold,linkage_group,chromosome',
 
             'IUPAC' : 'RYKMSWBDHV',    # symbols to be replaced with N in the DNA sequences (ensembl core(107) doesn't support the whole IUPAC alphabet for DNA)
 
@@ -109,6 +109,11 @@ class load_sequence_data(eHive.BaseRunnable):
 
             # loading additional sequences to the already exsisting core db
             'load_additional_sequences' : 0,
+
+            # size of the sequence data chunk, if 0 (default), no chunking is performed
+            'sequence_data_chunck' : 0,
+            # coord system name for chunks
+            'chunk_cs_name' : 'ensembl_internal',
         }
 
 
@@ -180,11 +185,10 @@ class load_sequence_data(eHive.BaseRunnable):
         # start coord system ranking and agps processing
         agps = self.from_param("manifest_data", "agp", not_throw = True)
 
-        # rank cs_names, met in agps.keys ("-" separated, i.e. "scaffold-contig") based on cs_order
+        # get the deafult coord_system order
         #   use noagp_cs_name_default for "noagp" assemblies
         cs_order = self.coord_sys_order(self.param("cs_order"))
         noagps_cs = self.param("noagp_cs_name_default")
-        cs_rank = self.used_cs_ranks(agps, cs_order, noagps_cs)
 
         # remove gaps and lower_level mappings if the are coveres by higher level ones
         #   i.e.: remove 'contigN to chromosomeZ', if 'contigN to scaffoldM' and 'scaffoldM to chromosomeZ' are in place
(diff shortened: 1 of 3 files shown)


lib/python/ensembl/brc4/runnable/load_sequence_data.py
```diff
@@ -791,25 +810,53 @@ def used_cs_ranks(self, agps, cs_order, noagp_default = None):
         return { e:i for i,e in enumerate(sorted(cs_used_set,key=lambda x:-cs_order[x]), start=1) }
 
 
+    def chunk_contigs(self, fasta, cs_ranks, agps, work_dir, chunk_size = 0, chunks_cs_name = "ensembl_internal"):
+        """
+        chunk dna sequence fasta
+          no chunking if chunk_size < 50k
+        """
+        if chunk_size < 50_000:
```
MatBarba, line 818: The min chunk size should probably be defined somewhere instead of being hard coded here

MatBarba:
What is the rational for this: is it to handle smaller DNA sequences in the database, for better performance?
Should this be applied to any new genome loaded?

lib/python/ensembl/brc4/runnable/load_sequence_data.py
```diff
@@ -791,25 +810,53 @@ def used_cs_ranks(self, agps, cs_order, noagp_default = None):
...
+            return fasta, cs_ranks, agps
+
+        # split using script
+        en_root = self.param_required("ensembl_root_dir")
+        _splitter = pj(en_root, r"ensembl-genomio/scripts/chunk_fasta.py")
```
MatBarba, line 823: It'd be a good idea to use paths here, I forgot what pj was supposed to be :)

vsitnik:
@MatBarba it's to allow us to load 1GB long contings into mysql using our current API.
This approach was used previously to speed things a bit on the web. But there's no need to so at the moment. They've fixed something on our side. So it's just to deal with our inability to load long contigs then to speed up anything

lib/python/ensembl/brc4/runnable/load_sequence_data.py
```diff
@@ -791,25 +810,53 @@ def used_cs_ranks(self, agps, cs_order, noagp_default = None):
...
+        print(f"running {split_cmd}", file = sys.stderr)
+        # NB throws CalledProcessError if failed  
+        sp.run(split_cmd, shell=True, check=True)
+
+        # add rank for chunks
+        _cs_name, _cs_rank = sorted(cs_ranks.items(), key=lambda k:k[1])[-1]
+        cs_ranks[chunks_cs_name] = _cs_rank + 1
+
+        # add agps entry
+        if agps is None: agps = dict()
```
MatBarba, line 841:
```suggestion
        if agps is None:
            agps = dict()
```

vsitnik: personal preferences :)

lib/python/ensembl/brc4/runnable/load_sequence_data.py
vsitnik: thanks!

lib/python/ensembl/brc4/runnable/load_sequence_data.py
vsitnik: path join

lib/python/ensembl/brc4/runnable/load_sequence_data.py
```diff
@@ -791,25 +810,53 @@ def used_cs_ranks(self, agps, cs_order, noagp_default = None):
...
+        # split using script
+        en_root = self.param_required("ensembl_root_dir")
+        _splitter = pj(en_root, r"ensembl-genomio/scripts/chunk_fasta.py")
+
+        os.makedirs(work_dir, exist_ok=True)
+
+        _stderr = f"{work_dir}/chunking.stderr"
+        _out_agp = f"{work_dir}/chunks.agp"
+        _out_fasta = f"{work_dir}/chunks.fasta"
+        split_cmd = f"python {_splitter} --chunk_size {chunk_size} --agp_out {_out_agp} --out {_out_fasta} {fasta} 2> {_stderr}"
```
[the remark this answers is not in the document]
vsitnik: safer, as for me. and the log is stored closer to the data for this step

MatBarba:
Another note (not blocking): the script file is in the folder /scripts, but it's part of a pipeline, so not a script per se (=that should be run separately).
We'll need to better organize the files for e.g. when we move to nextflow.

lib/python/ensembl/brc4/runnable/load_sequence_data.py
vsitnik: done. "sequence_data_chunck_min_len" parameter added.

lib/python/ensembl/brc4/runnable/load_sequence_data.py
vsitnik: Will be done with a separate PR.

PR #34, opened 2022-12-16
Mbarba/mobile genetic element

MatBarba:
Convert mobile_genetic_element features into transposable_element ones.

Keep the mobile_element_type in the description.

### DIFF
--- a/lib/python/ensembl/brc4/runnable/integrity.py
+++ b/lib/python/ensembl/brc4/runnable/integrity.py
@@ -159,6 +159,7 @@ class integrity(eHive.BaseRunnable):
                     if len(tr_errors) > 0:
                         tr_errors = self.check_ids(func_ann["translations"], gff["all_translations"], "Translation ids metadata vs gff (include pseudo CDS)")
                     errors += tr_errors   
+                    errors += self.check_ids(func_ann["transposable_elements"], gff["transposable_elements"], "TE ids metadata vs gff")
 
                 #Check the seq.json intregrity
                 #Compare the length and id retrieved from seq.json to the gff
@@ -262,14 +263,17 @@ class integrity(eHive.BaseRunnable):
             # Get gene ids and translation ids
             genes = {}
             translations = {}
+            tes = {}
 
             for item in data:
                 if item["object_type"] == "gene":
                     genes[item["id"]] = 1
                 elif item["object_type"] == "translation":
                     translations[item["id"]] = 1
+                if item["object_type"] == "transposable_element":
+                    tes[item["id"]] = 1
 
-            return { "genes" : genes, "translations" : translations }
+            return { "genes" : genes, "translations" : translations, "transposable_elements": tes }
 
     def get_gff3(self, gff3_path):
         #Load the gff file
@@ -297,17 +301,19 @@ class integrity(eHive.BaseRunnable):
         genes = {}
         peps = {}
         all_peps = {}
+        tes = {}
 
         gff = GFF.parse(gff3_handle)
         for seq in gff:
             seqs[seq.id] = len(seq.seq)
             
             for feat in seq.features:
+                feat_length = abs(feat.location.end - feat.location.start)
                 #Store gene id and length
-                if feat.type in ["gene", "ncRNA_gene", "pseudogene", "transposable_element"]:
+                if feat.type in ["gene", "ncRNA_gene", "pseudogene"]:
                     gene_id = feat.id
                     if ensembl_mode:
-                        gene_id = gene_id.replace("gene:", "")
+                        gene_id = feat_length
                     genes[gene_id] = abs(feat.location.end - feat.location.start)
                     # Get CDS id and length
                     for feat2 in feat.sub_features:
@@ -327,9 +333,17 @@ class integrity(eHive.BaseRunnable):
                                 if feat.type != "pseudogene":
                                     peps[pep_id] = pep_length
                                 all_peps[pep_id] = pep_length
+                if feat.type == 'transposable_element':
+                    tes[feat.id] = feat_length
 
-
-        return { "seq_region": seqs, "genes": genes, "translations": peps, "all_translations": all_peps }
+        stats = {
+            "seq_region": seqs,
+            "genes": genes,
+            "translations": peps,
+            "all_translations": all_peps,
+            "transposable_elements": tes    
+        }
+        return stats
 
     def get_agp_seq_regions(self, agp_dict):
         """AGP files describe the assembly of larger sequence objects using smaller objects.
--- a/lib/python/ensembl/brc4/runnable/process_gff3.py
+++ b/lib/python/ensembl/brc4/runnable/process_gff3.py
@@ -288,7 +288,7 @@ class process_gff3(eHive.BaseRunnable):
             mobile_element_type = feat.qualifiers.get("mobile_element_type", [])
             if mobile_element_type:
                 # Get the type (and name) from the attrib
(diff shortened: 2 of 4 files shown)


lib/python/ensembl/brc4/runnable/integrity.py
```diff
@@ -297,17 +301,19 @@ def parse_gff3(self, gff3_handle):
...
         peps = {}
         all_peps = {}
+        tes = {}
 
         gff = GFF.parse(gff3_handle)
         for seq in gff:
             seqs[seq.id] = len(seq.seq)
             
             for feat in seq.features:
+                feat_length = abs(feat.location.end - feat.location.start)
```
JAlvarezJarreta, line 311: "Important" question: are the start and end supposed to be both part of the feature? If so, a `+1` would be required to get the exact length (I have always despised these random decisions on whether `end` is actually the last character or the next one :unamused:)

lib/python/ensembl/brc4/runnable/process_gff3.py
```diff
@@ -264,16 +270,44 @@ def simpler_gff3(self, gff3_in: TextIO, out_gff_path: Path, out_funcann_path: Pa
...
+        # Also generate an ID if needed
+        feat.id = self.normalize_gene_id(feat)
+        self.add_funcann_feature(functional_annotation, feat, "transposable_element")
+        feat.qualifiers = {
+            "ID": feat.id
+        }
+
+        return feat
+
+    def normalize_gene(self, gene, functional_annotation, fail_types):
```
JAlvarezJarreta, line 310:
```suggestion
    def normalize_gene(self, gene: SeqFeature, functional_annotation: List, fail_types: List) -> SeqFeature:
```

lib/python/ensembl/brc4/runnable/process_gff3.py
```diff
@@ -264,16 +270,44 @@ def simpler_gff3(self, gff3_in: TextIO, out_gff_path: Path, out_funcann_path: Pa
...
-    def normalize_gene(self, gene: SeqFeature, functional_annotation: List, fail_types: List) -> SeqFeature:
+    def format_mobile_element(self, feat, functional_annotation):
+        """Given a mobile_genetic_element feature, transform it into a transposable_element"""
+        quals = feat.qualifiers
+
+        if feat.type == 'mobile_genetic_element':
+            mobile_element_type = feat.qualifiers.get("mobile_element_type", [])
+            if mobile_element_type:
+                element_type, transposon_type = mobile_element_type[0].split(':')
+                if element_type in ("transposon", "retrotransposon"):
```
JAlvarezJarreta, line 289:
```suggestion
                if element_type.endswith("transposon"):
```

JAlvarezJarreta (changes requested):
Nice! And it seems the amount of code required was quite minimal, which is always great news. Only the last comment is something that should be done.

lib/python/ensembl/brc4/runnable/process_gff3.py
MatBarba: I'd prefer we keep the explicit list, since those are ontology terms (which we might expand in the future if for some reason there are other types or transposons we might accept)

lib/python/ensembl/brc4/runnable/integrity.py
MatBarba: This length is compared with the one calculated from the peptide fasta length, so the comparison should have raised errors quickly if we didn't... assuming the code that compares the lengths does so correctly.
I will need to review that part of the code to make sure (a good example of why we need some tests); though that's not the purpose of this PR.

JAlvarezJarreta (approved):

vsitnik (approved):
ok, as for me. I'll have to add some changes on the mz side of things.

lib/python/ensembl/brc4/runnable/process_gff3.py
```diff
@@ -264,15 +270,43 @@ def simpler_gff3(self, gff3_in: TextIO, out_gff_path: Path, out_funcann_path: Pa
...
+                    feat.type = 'transposable_element'
+                    if not feat.qualifiers.get("product"):
```
vsitnik, line 291: could you post some examples here? Thanks

MatBarba: From GCA_002219185.1 (C. glabrata DSY562):
`CM008085.1      Genbank mobile_genetic_element  1575    7277    .       +       .       ID=id-CM008085.1:1575..7277;Note=label: mobile_element;gbkey=mobile_element;mobile_element_type=retrotransposon;rpt_type=long_terminal_repeat`

From GCA_026262505.1 (T. vaginalis):
`CM044617.1      Genbank mobile_genetic_element  299     2423    .       -       .       ID=id-CM044617.1:299..2423;gbkey=mobile_element;mobile_element_type=transposon:rnd-1_family-319/TvMULE1`

(Note how the first has only the element type, and the second also has a name)

lib/python/ensembl/brc4/runnable/process_gff3.py
```diff
@@ -37,6 +37,9 @@ def param_defaults(self):
                 "pseudogene",
                 "ncRNA_gene",
             ),
+            "non_gene_types": (
```
JAlvarezJarreta, line 40: I'm not 100% happy with this key, but I cannot think of an alternative that sounds better to me.

lib/python/ensembl/brc4/runnable/process_gff3.py
```diff
@@ -248,13 +252,15 @@ def simpler_gff3(self, gff3_in: TextIO, out_gff_path: Path, out_funcann_path: Pa
...
                     elif feat.type == "CDS":
                         feat = self.cds_gene(feat)
+                    elif feat.type in ('mobile_genetic_element', 'transposable_element'):
+                        feat = self.format_mobile_element(feat, functional_annotation)
                     
                     # Normalize the gene structure
                     if feat.type in allowed_gene_types:
                         feat = self.normalize_gene(feat, functional_annotation, fail_types)
-                    else:
+                    elif feat.type not in allowed_non_gene_types:
```
JAlvarezJarreta, line 261: If we can think of a future expansion of the `allowed_non_gene_types`, we can go with:
```suggestion
                    elif feat.type in allowed_non_gene_types:
                        pass
                    else:
```

JAlvarezJarreta (changes requested):

lib/python/ensembl/brc4/runnable/process_gff3.py
```diff
@@ -264,15 +270,53 @@ def simpler_gff3(self, gff3_in: TextIO, out_gff_path: Path, out_funcann_path: Pa
...
+                    feat.type = 'transposable_element'
+                    if not feat.qualifiers.get("product"):
+                        feat.qualifiers["product"] = [description]
+                else:    
+                    print(f"Mobile genetic element 'mobile_element_type' is not transposon: {element_type}")
+            else:
+                print("Mobile genetic element does not have a 'mobile_element_type' tag")
+        
+        # At this point, if it is valid it should be a transposable_element feature
+        if feat.type != "transposable_element":
```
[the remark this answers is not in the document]
MatBarba: This if assumed that some mobile_genetic_elements would not be changed to transposable_element (because the format would not be as expected), and so would return here, before adding things to the functional annotation.

I've changed the code above to return the feature directly, so that elif is now appropriate.

JAlvarezJarreta (approved):
Nice to have this addition with so little changes.

PR #39, opened 2023-01-31
Mbarba/stats

MatBarba:
Use the program "datasets" from ncbi (if available in the PATH) to get stats for each genome annotation, and compare the number of each biotype with what we have in the final gff to be loaded.
The stats are written to the stats file.
The runnable will fail if there are differences, in which case check why it differs (might be a parser bug, or wrong counts on NCBI side e.g. MT genes not counted correctly). If the differences are accounted for, just forgive the analyses.


lib/python/ensembl/brc4/runnable/manifest_stats.py
```diff
@@ -168,13 +195,68 @@ def parse_gff3(self, gff3_handle: TextIO) -> List:
...
         
         return stats
+    
+    def check_ncbi_stats(self, biotypes: Dict, accession: str) -> List:
+        """Use the dataset tool from NCBI to get stats and compare with what we have"""
+
+        stats = []
+
+        datasets_bin = self.param('datasets_path')
+        if not which(datasets_bin):
```
JAlvarezJarreta, line 210: This "conversion" of `datasets_path` to `datasets_bin` makes the code a little bit confusing: the former gives me the impression it is a path to a folder with dataset files, and the latter that the path must be to a binary that process datasets? Not sure if applicable in this PR, but it would be good to revisit this naming later on.

lib/python/ensembl/brc4/runnable/manifest_stats.py
```diff
@@ -168,13 +195,68 @@ def parse_gff3(self, gff3_handle: TextIO) -> List:
...
+
+        # Get stats
+        if "reports" in result:
+            genome = result["reports"][0]
+            if "annotation_info" in genome and "stats" in genome["annotation_info"]:
+                ncbi_stats = genome["annotation_info"]["stats"]
+
+                if "gene_counts" in ncbi_stats:
+                    counts = ncbi_stats["gene_counts"]
+                    stats += self.compare_ncbi_counts(biotypes, counts)
```
JAlvarezJarreta, line 226: Since the list should be empty at this point, I believe you can get away with less "compute":
```suggestion
                    stats = self.compare_ncbi_counts(biotypes, counts)
```

JAlvarezJarreta (approved):
It looks pretty straight-forward to me, just a couple of minor edits and suggestions.

The only thing we need to do once this is rolled out is to make it clear somewhere that the stats we are providing are not NCBI's but ours (because I believe UX specifically mentioned NCBI ones).

MatBarba:
> 

Note that those stats are for out own use only, not for handover. They are intermediate stats for the files that we prepare before loading.
Handover stats will have to be done separately, in the dumper pipeline I think (though the code will be very similar and easy to write from this).

PR #42, opened 2023-02-14
Mbarba/dump events

MatBarba:
Integrate the stable ids event dump in the dumper pipeline. The ids history will be dumped automatically if there is any, in the BRC4 expected format.
Also if there is no files created, then don't create a manifest or folder (e.g. when dumping history only, and some dbs don't have any).

### DIFF
--- a/lib/perl/Bio/EnsEMBL/Pipeline/Runnable/BRC4/Manifest.pm
+++ b/lib/perl/Bio/EnsEMBL/Pipeline/Runnable/BRC4/Manifest.pm
@@ -35,9 +35,13 @@ use v5.14;
 sub run {
   my ($self) = @_;
 
-  my $manifest = $self->param_required('manifest');
-  my $output_dir = $self->param_required('output_dir');
   my $species = $self->param_required('species');
+  my $manifest = $self->param('manifest');
+  if (not manifest_has_files($manifest)) {
+    say("No manifest data for $species");
+    return;
+  }
+  my $output_dir = $self->param_required('output_dir');
   my $species_component = $self->get_meta_value('BRC4.component');
   my $species_abbrev = $self->get_meta_value('BRC4.organism_abbrev');
   my $species_name = $species_abbrev // $species;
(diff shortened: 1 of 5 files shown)


lib/perl/Bio/EnsEMBL/Pipeline/Runnable/BRC4/Manifest.pm
```diff
@@ -75,6 +79,26 @@ sub run {
...
 }
 
+sub manifest_has_files {
+  my ($manifest) = @_;
+
+  my $has_files = 0;
+  for my $name (keys %$manifest) {
+
+    if (ref($manifest->{$name}) eq 'HASH') {
+      foreach my $subname (sort keys %{ $manifest->{$name} }) {
```
JAlvarezJarreta, line 89: Why `sort`? I do not see the value on sorting the keys.

lib/python/ensembl/brc4/runnable/dump_stable_ids.py
```diff
@@ -215,6 +216,19 @@ def _name_event(self) -> None:
...
+        """Remove the empty old pairs when the event is not 'new'."""
+        if not self.name:
+            self._name_event()
+        
+        if self.name != "new":
+            new_pairs = []
+            for pair in self.pairs:
+                if pair["old_id"] == "":
+                    continue
+                new_pairs.append(pair)
```
JAlvarezJarreta, line 229: This is the only way I can think that justifies the need of `continue`:
```suggestion
                if ("old_id" in pair) and (pair["old_id"] == ""):
                    continue
                new_pairs.append(pair)
```

schema/manifest_schema.json
```diff
@@ -15,7 +15,8 @@
         "seq_attrib" : { "$ref" : "#/definitions/file_info" },
         "genome" : { "$ref" : "#/definitions/file_info" },
         "functional_annotation" : { "$ref" : "#/definitions/file_info" },
-        "agp" : { "$ref" : "#/definitions/agp_info" }
+        "agp" : { "$ref" : "#/definitions/agp_info" },
+        "events" : { "$ref" : "#/definitions/file_info" }
```
JAlvarezJarreta, line 19: This is the most annoying thing for me from JSON: not being able to finish the last element with `,` so the new additions only cover new lines.

JAlvarezJarreta (approved):
I find it surprising that there is no runnable like `ExtractConnection.pm` in Hive :exploding_head: 

_Note:_ -1 point for adding a new Perl file to the codebase though :stuck_out_tongue_winking_eye:

lib/python/ensembl/brc4/runnable/dump_stable_ids.py
MatBarba: Replaced with .get()

schema/manifest_schema.json
MatBarba: Good idea to add a json checker to our CI/CD then?

JAlvarezJarreta: Yes, this is already planned once I finally sit down and push on the CI/CD documentation.

PR #44, opened 2023-02-28
Update chunk_fasta.py

digrigor:
The "if split_points and" in L74 was preventing the script to add the seq_len to the end of the list if there were no Ns in the sequence. This is fixed by removing it.


vsitnik (approved):
thanks!

PR #45, opened 2023-03-01
Fix BRC4/EBI seq region names

MatBarba:
Use the list of sources in order (so GenBank first). For EBI, use whatever is used for the name.


lib/python/ensembl/brc4/runnable/process_seq_region.py
```diff
@@ -190,34 +189,31 @@ def add_brc4_ebi_name(self, seq_regions: List[SeqRegion]) -> List[SeqRegion]:
...
-            else:
+            # Choose the synonym to use as the BRC name
+            # Use the first valid name from the source list
+            brc_name = ''
+            for source_name in sources:
+                if source_name in names:
+                    brc_name = names[source_name]
+                    break
+            
+            if not brc_name:
```
JAlvarezJarreta, line 208: This is an alternative implementation (Python's tricks!):
```suggestion
            for source_name in sources:
                if source_name in names:
                    brc_name = names[source_name]
                    break
            else:
```
Source: https://www.w3schools.com/python/gloss_python_for_else.asp

lib/python/ensembl/brc4/runnable/process_seq_region.py
```diff
@@ -190,34 +189,31 @@ def add_brc4_ebi_name(self, seq_regions: List[SeqRegion]) -> List[SeqRegion]:
...
                 raise Exception(f"Can't set BRC4/EBI name for {seqr}")
             
-            parts = name.partition(".")
-            flat_name = parts[0]
-            seqr["BRC4_seq_region_name"] = flat_name
-            seqr["EBI_seq_region_name"] = flat_name
+            brc_name = brc_name.partition(".")[0]
```
JAlvarezJarreta, line 211: Unless you plan to add elements with `.` in `brc4_synonym_sources` in the future or I'm missing something, I do not believe you should ever encounter a `.` inside `brc_name`, right?

JAlvarezJarreta (approved):

lib/python/ensembl/brc4/runnable/process_seq_region.py
MatBarba: brc_name contains either a Genbank or a RefSeq accession, both usually contain a . version. Hence the need to remove the . version here

PR #55, opened 2023-03-15
Run `black` over the whole repository

JAlvarezJarreta:
As a team we have agreed to apply `black` to our codebase to ensure we are PEP8 compliant. In the future this will be run automatically during code submission to the repository, so you will not need to run it "manually" (and won't forget).

### DIFF
--- a/scripts/brc4/gff-overlaps.py
+++ b/scripts/brc4/gff-overlaps.py
@@ -24,138 +24,166 @@ import logging
 import json
 from collections import defaultdict
 
-def stats( in_file ):
+
+def stats(in_file):
     """run analysis of GFF file and produce a summary of feature types"""
 
     examiner = GFFExaminer()
     in_handle = open(in_file)
 
-    print( f"\nrunning analysis of GFF file\n" )
-    
+    print(f"\nrunning analysis of GFF file\n")
+
     pprint.pprint(examiner.available_limits(in_handle))
-    print("\n\n" )
+    print("\n\n")
     in_handle.close()
     sys.exit(0)
-    
-def scan_tree( intervals ):
+
+
+def scan_tree(intervals):
     """construct an interval tree using supplied genomic intervals, check all elements on the tree against iself and return any that hit 2 or more intervals (i.e. itself + 1 other)"""
 
     retlist = set()
     t = IntervalTree(Interval(*iv) for iv in intervals)
-    
+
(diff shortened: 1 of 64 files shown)


ens-LCampbell (approved):

src/python/ensembl/brc4/runnable/extract_from_gb.py
```diff
@@ -143,117 +141,109 @@ def _get_organella(self, gb_file):
...
-                                )
+                            peptides.append(SeqRecord(Seq(feat.qualifiers["translation"][0]), id=cds_id))
 
                             # Also create a parent transcript for this translation
-                            tr_qualifiers = {
-                                    "ID" : tr_id,
-                                    "Name" : gene_name,
-                                    "Parent": gene_id
-                                    }
+                            tr_qualifiers = {"ID": tr_id, "Name": gene_name, "Parent": gene_id}
```
MatBarba, line 195: To keep this in multiline, add a comma after the last element

vsitnik (approved):
ok for me. thanks!

scripts/brc4/gff-overlaps.py
```diff
@@ -24,138 +24,166 @@
...
+        required=True,
+        nargs="?",
+        default="overlaps",
+    )
+    parser.add_argument(
+        "--filter",
+        help="seqfeature type to use for overlap isolation (default=gene)",
+        nargs="?",
+        default="gene",
+    )
```
MatBarba, line 174: It's a bit weird but that's what the line limit does... I would actually be ok with them all being multiline, at least that'd be uniform (we can force that by adding a comma after the last element, cf last 2 add_argument)

MatBarba (approved):
Looks ok to me. We can only force some things when we come across them (multiline collections).

scripts/brc4/gff-overlaps.py
JAlvarezJarreta: In my opinion it may make some files significantly longer, but I do not mind either way.

JAlvarezJarreta:
> Don't you remember we used some option to make strings a bit longer? Personally, I'm not a big fan of `parser.add_argument` argument list being split.

I'm afraid 110 is as long as we can go based on the agreement Ensembl-wide.

Dishalodha (approved):

PR #56, opened 2023-03-15
Fix BRC4/EBI seq region names

MatBarba:
Use the list of sources in order (so GenBank first). For EBI, use whatever is used for the name.


MatBarba:
NB: remade the PR rebased on main after the tree change and black formatting

PR #70, opened 2023-04-05
Disha/nextflow add seq

MatBarba:
Final PR to merge the nextflow pipelines in development, so that we can work more closely to the main branch.

This includes:
- the addition of the pipelines/nextflow folder with both additional_seq_prepare and dumper_pipeline
- The conversion of Python modules to replace the runnables from eHive and to be used in the nextflow pipelines. Those modules are accessed by the pipelines via the environment defined in pyproject.toml.
- A starting set of tests (src/python/tests)


pyproject.toml
```diff
@@ -53,46 +44,80 @@ classifiers = [
...
     "mock",
     "Sphinx",
 ]
 
-
 [project.urls]
 homepage = "https://www.ensembl.org"
 repository = "https://github.com/Ensembl/ensembl-genomio"
 
+[project.scripts]
```
vsitnik, line 74: mb add some comments to allow for grouping

src/python/ensembl/brc4/runnable/core_server.py
```diff
@@ -88,14 +88,56 @@ def get_cores(self, prefix: str = "", build: str = "", version: str = "") -> Lis
         dbs = self.get_all_cores()
 
         if prefix:
-            dbs = [db for db in dbs if db.startswith(f"{prefix}_")]
+            dbs = [db for db in dbs if db.startswith(f"{prefix}")]
```
vsitnik, line 91: why so?

src/python/ensembl/io/genomio/download_genbank.py
vsitnik, line 1: is a duplicate of src/python/ensembl/brc4/runnable/download_genbank.py

vsitnik (commented):
Can we resolve duplication between
src/python/ensembl/brc4/runnable/
and
src/python/ensembl/io/genomio
?
When would be able to do so?

src/python/ensembl/brc4/runnable/core_server.py
MatBarba: It's a bit more flexible, think more along the line of selecting databases with their "start" than their prefix. An example: if I create databases with several prefixes like mbarba64a_, mbarba64b_ etc. I will still be able to get the whole list with a prefix="mbarba64".

README.md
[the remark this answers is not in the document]
MatBarba: Yes

src/python/ensembl/brc4/runnable/core_server.py
vsitnik: cool. thanks!

src/python/ensembl/io/genomio/download_genbank.py
vsitnik: 😭

PR #74, opened 2023-04-28
Mbarba/events loader2

MatBarba:
An independent module to load events in a core database.

The current version load events as they can be dumped by the event_dumper module (with a BRC export).

### DIFF
--- a/src/python/ensembl/io/genomio/events_loader.py
+++ b/src/python/ensembl/io/genomio/events_loader.py
@@ -28,6 +28,7 @@ from ensembl.core.models import MappingSession, StableIdEvent
 
 @dataclass
 class IdEvent:
+    """Simple representation for the events from the input file"""
     from_id: str
     to_id: str
     release: str
@@ -35,16 +36,22 @@ class IdEvent:
 
 
 class MapSession:
+    """Simple mapping_sessions representation from the input file"""
     def __init__(self, release: str, release_date: str) -> None:
         self.release = release
         self.release_date = release_date
         self.events: List[IdEvent] = []
 
     def add_event(self, event: IdEvent) -> None:
+        """Add an event to this mapping_session"""
         self.events.append(event)
 
 
 def load_events(input_file: Path) -> List[IdEvent]:
+    """Load events from input file.
+    Expected tab file columns: old_id, new_id, event_name, release, release_date
+    
+    """
     events: List[IdEvent] = []
 
     with input_file.open("r") as events_fh:
@@ -59,6 +66,10 @@ def load_events(input_file: Path) -> List[IdEvent]:
 
 
 def write_events(session: Session, events: List[IdEvent], update: bool = False) -> None:
+    """Insert the events in the core database.
+    A mapping session is created for each different 'release'.
+    
+    """
     # First, create mapping_sessions based on the release
     mappings: Dict[str, MapSession] = {}
     for event in events:
@@ -116,6 +127,7 @@ class InputSchema(argschema.ArgSchema):
 
 
 def make_mysql_url(host: str, user: str, database: str, port: int = 0, password: str = "") -> str:
+    """Create a mysql URL given server parameters for SQLalchemy."""
     user_pass = user
     host_port = host
     if password:
--- a/src/python/ensembl/io/genomio/events_loader.py
+++ b/src/python/ensembl/io/genomio/events_loader.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+# See the NOTICE file distributed with this work for additional information
+# regarding copyright ownership.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


src/python/ensembl/io/genomio/events_loader.py
```diff
@@ -0,0 +1,166 @@
...
+        release = event.release
+
+        if release in mappings:
+            mappings[release].add_event(event)
+        else:
+            mappings[release] = MapSession(release, event.release_date)
+
+    # Then, add the mapping, and the events for this mapping
+    for release, mapping in mappings.items():
+        print(f"Adding mapping for release {release} ({len(mapping.events)} events)")
```
JAlvarezJarreta, line 85: Just a more generic question: should we consider moving from `print` messages to a proper log file?

src/python/ensembl/io/genomio/events_loader.py
```diff
@@ -0,0 +1,166 @@
...
+                session.add(id_event)
+            session.commit()
+
+
+class InputSchema(argschema.ArgSchema):
+    """Input arguments expected by this script."""
+
+    # Server parameters
+    host = argschema.fields.String(
+        required=True, metadata={"description": "Host to the server with EnsEMBL databases"}
```
JAlvarezJarreta, line 109: I have also seen previously the option to write this as:
```
metadata={"required": True, "description": "Host to the server with EnsEMBL databases"}
```
In my opinion, this is something that should be fixed in `argschema` so `description` can go again outside of `metadata`, but I'm afraid it won't happen any time soon.

src/python/ensembl/io/genomio/events_loader.py
```diff
@@ -0,0 +1,166 @@
...
+def make_mysql_url(host: str, user: str, database: str, port: int = 0, password: str = "") -> str:
+    """Create a mysql URL given server parameters for SQLalchemy."""
+    user_pass = user
+    host_port = host
+    if password:
+        user_pass = f"{user}:{password}"
+    if port:
+        host_port = f"{host}:{port}"
+    db_url = f"mysql://{user_pass}@{host_port}/{database}"
+    return db_url
```
JAlvarezJarreta, line 138: You can use `sqlalchemy` for this (see below).

src/python/ensembl/io/genomio/events_loader.py
```diff
@@ -0,0 +1,166 @@
...
+    user = mod.args["user"]
+    password = mod.args.get("password")
+    database = mod.args.get("database")
+    db_url = make_mysql_url(
+        host=host,
+        port=port,
+        user=user,
+        password=password,
+        database=database,
+    )
```
JAlvarezJarreta, line 156:
```
from sqlalchemy.engine import URL
...
db_url = URL.create("mysql", mod.args["user"], mod.args.get("password"), mod.args["host"], mod.args["port"], mod.args.get("database"))
```
_Note:_ No need to `int()` port, since it will be already an integer.

src/python/ensembl/io/genomio/events_loader.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def load_events(input_file: Path) -> List[IdEvent]:
+    """Load events from input file.
+    Expected tab file columns: old_id, new_id, event_name, release, release_date
+    
+    """
+    events: List[IdEvent] = []
+
+    with input_file.open("r") as events_fh:
```
JAlvarezJarreta, line 57: I believe it is better to extend this method:
```suggestion
def load_events(input_file: PathLike) -> List[IdEvent]:
    """Load events from input file.
    Expected tab file columns: old_id, new_id, event_name, release, release_date
    
    """
    events: List[IdEvent] = []

    with Path(input_file).open("r") as events_fh:
```
_Note:_ you will need to add `from os import PathLike` above.

src/python/ensembl/io/genomio/events_loader.py
```diff
@@ -0,0 +1,166 @@
...
+    db_url = make_mysql_url(
+        host=host,
+        port=port,
+        user=user,
+        password=password,
+        database=database,
+    )
+    dbc = DBConnection(db_url)
+
+    events = load_events(Path(mod.args.get("input_file")))
```
JAlvarezJarreta, line 159: If you follow my suggestion above regarding `PathLike`:
```suggestion
    events = load_events(mod.args.get("input_file"))
```

src/python/ensembl/io/genomio/events_loader.py
```diff
@@ -0,0 +1,166 @@
...
+
+    with dbc.session_scope() as session:
+        write_events(session, events, mod.args["update"])
+
+
+if __name__ == "__main__":
+    main()
```
JAlvarezJarreta, line 166: I would rather add an entry-point to `pyproject.toml` and remove this.

JAlvarezJarreta (changes requested):
A few suggestions and a couple of minor edits. Good to get the chance to learn something new in a review :nerd_face:

src/python/ensembl/io/genomio/events_loader.py
MatBarba: Yes we should...

src/python/ensembl/io/genomio/events_loader.py
MatBarba: It doesn't seem to work properly when it is in metadata (and the only way to tell is by running the script).

src/python/ensembl/io/genomio/events_loader.py
MatBarba: I've added an entrypoint, though I'd still prefer we keep this part for dev testing

PR #78, opened 2023-05-10
Mbarba/prepare addition fix

MatBarba:
Fixes for the addition prepare hive pipeline

PR #87, opened 2023-05-22
replicating the changes made by Matthieu in nextflow

Dishalodha:
Replicating the changes in nextflow to use locus_tag as gene name when required and adding support for pseudogene in the gff while adding an organellar sequence. 

Derived from this commit by Matthieu
https://github.com/Ensembl/ensembl-genomio/pull/78

### DIFF
--- a/src/python/ensembl/io/genomio/download_genbank.py
+++ b/src/python/ensembl/io/genomio/download_genbank.py
@@ -44,7 +44,7 @@ def download_genbank(accession: str) -> str:
     e_url = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi"
     e_params = {
         "db": "nuccore",
-        "rettype": "gb",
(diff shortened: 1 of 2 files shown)


src/python/ensembl/io/genomio/download_genbank.py
```diff
@@ -44,7 +44,7 @@ def download_genbank(accession: str) -> str:
     e_url = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi"
     e_params = {
         "db": "nuccore",
-        "rettype": "gb",
+        "rettype" : "gbwithparts",
```
JAlvarezJarreta, line 47:
```suggestion
        "rettype": "gbwithparts",
```

JAlvarezJarreta (changes requested):
PEP8 for the win!

PR #92, opened 2023-05-24
NOTICE copyright year check

JAlvarezJarreta:
Added new CI/CD pipeline for License-related checks, starting for checking if the NOTICE file has the latest copyright year.

I have also taken the opportunity to do a small change in the coverage check to use the library namespace instead of the source, so if GenomIO is not installed with `-e` it should still return the right coverage information.


MatBarba (approved):
Looks good, thanks!

PR #93, opened 2023-05-24
Mbarba/cleanup

MatBarba:
Tweaks and fixes to make mypy happy

### DIFF
--- a/src/python/ensembl/io/genomio/utils/archive_utils.py
+++ b/src/python/ensembl/io/genomio/utils/archive_utils.py
@@ -38,8 +38,9 @@ def extract_file(src_file: PathLike, dst_dir: PathLike) -> None:
         dst_dir: Path to where extract the file.
 
     """
-    extension = Path(src_file).suffix
-    file_base = src_file.strip(extension)
+    src_file = Path(src_file)
+    extension = src_file.suffix
+    file_base = src_file.stem
     final_path = Path(dst_dir) / file_base
 
     if extension == ".gz":
--- a/src/python/ensembl/io/genomio/utils/archive_utils.py
+++ b/src/python/ensembl/io/genomio/utils/archive_utils.py
@@ -16,12 +16,25 @@
 
 __all__ = ["SUPPORTED_ARCHIVE_FORMATS", "extract_file"]
 
+from contextlib import contextmanager
 import gzip
 from os import PathLike
 from pathlib import Path
 import shutil
+from typing import Generator, TextIO
 
 import argschema
+
+
+@contextmanager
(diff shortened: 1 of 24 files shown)


src/python/ensembl/io/genomio/utils/archive_utils.py
```diff
@@ -16,14 +16,27 @@
...
 
 
+@contextmanager
+def open_gz_file(file_path: PathLike) -> Generator[TextIO, None, None]:
+    this_file = Path(file_path)
+    if this_file.suffix == ".gz":
+        with gzip.open(this_file, "rt") as fh:
+            yield fh
+    else:
+        with open(this_file, "rt") as fh:
```
JAlvarezJarreta, line 36:
```suggestion
        with this_file.open("rt") as fh:
```
To follow with other places in the codebase where it is used this way with `Path` variables.

src/python/tests/metadata/test_prepare_genome.py
```diff
@@ -35,10 +35,13 @@
 
 
 class TestSchemas:
+    test_data_dir = Path()
+    tmp_dir = Path()
```
JAlvarezJarreta, line 39:
```suggestion
    test_data_dir: Path
    tmp_dir: Path
```
Although not 100%, it seems based on the information I have been able to find that this should be ok :crossed_fingers:

src/python/tests/test_schemas.py
```diff
@@ -36,11 +36,15 @@
 
 
 class TestSchemas:
+    test_data_dir = Path()
+    schema_dir = Path()
+    tmp_dir = Path()
```
JAlvarezJarreta, line 41:
```suggestion
    test_data_dir: Path
    schema_dir: Path
    tmp_dir: Path
```
Same as before..

JAlvarezJarreta (approved):
Great work! I know it is not something too enjoyable to make `mypy` or `pylint` happy, but at least it is worth the effort in the long run.

Just three minor comments, two of which are more a matter of preference than anything else (the last two).

vsitnik (approved):
cool. thanks!

src/python/ensembl/io/genomio/utils/archive_utils.py
```diff
@@ -16,14 +16,27 @@
...
 from os import PathLike
 from pathlib import Path
 import shutil
+from typing import Generator, TextIO
 
 import argschema
 
 
+@contextmanager
+def open_gz_file(file_path: PathLike) -> Generator[TextIO, None, None]:
```
vsitnik, line 30: are we sure we don't want to pass additional args in `*args` and  have `"rt"`  hardcoded?

PR #97, opened 2023-05-30
Migrate license check from Bash to Python

JAlvarezJarreta:
Adding template files for both the NOTICE file and the license header.

I have opted for a single script that can do both tasks independently ('xor' mode). I have left both tasks in two separate jobs at least for the first run of CI/CD, to see how it goes, but happy to merge them into a single task as discussed earlier.

_Note:_ I have opted to use `argparse` instead of `argschema` to be able to use the "raw" Python Docker image.


cicd/gitlab/templates/license_header.tpl
vsitnik, line 1: Do we want to put licence related templates  to the repo root or something like `/license_templates/`?

cicd/gitlab/scripts/check_license.py
vsitnik, line 1: what about putting it into `/cicd/scripts` ?

vsitnik (approved):

cicd/gitlab/templates/license_header.tpl
JAlvarezJarreta: Since these are for the CI/CD bit only, that is why I left them there. Furthermore, maybe more templates will be added in the future, so that is why I left the folder name more "open". Happy to move it somewhere else if you prefer your option.

cicd/gitlab/scripts/check_license.py
JAlvarezJarreta: Ah sure! That makes sense :+1:

vsitnik: cool. thanks!

vsitnik:
cool. thanks.

cicd/scripts/check_license.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+    )
+    subparsers = parser.add_subparsers(title="License aspect to check", required=True, dest="{notice,header}")
+    parser_notice = subparsers.add_parser("notice", help="Check NOTICE file format and copyright year")
+    parser_notice.set_defaults(func=check_notice)
+    parser_header = subparsers.add_parser("header", help="Check if code files have proper license header")
+    parser_header.set_defaults(func=check_header)
+    args = parser.parse_args()
+    # Run the corresponding function depending on the type of check selected by the user
+    args.func()
```
vsitnik, line 103: nice trick

vsitnik (approved):

cicd/scripts/check_license.py
```diff
@@ -0,0 +1,103 @@
...
+from pathlib import Path
+import re
+import sys
+
+
+# Do not show traceback for an easier error readability
+sys.tracebacklimit = 0
+
+
+_SCRIPTS_DIR = Path(__file__).absolute().parent
```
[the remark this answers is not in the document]
JAlvarezJarreta: I'm afraid I do not understand what do you mean. Add arguments to be able to pass different template paths?

MatBarba (commented):
A couple things to fix/explain, otherwise that looks good

cicd/scripts/check_license.py
```diff
@@ -0,0 +1,103 @@
...
+    template = "[#/-]*\s*" + template.replace("\n\n", "(\n)+").replace("\n", "\n[#/-]*\s*")
+    # Compile template as regex to improve performance
+    prog = re.compile(rf"{template}")
+    report = ""
+    for file_path in _ROOT_PATH.rglob("*.*"):
+        if file_path.is_file() and (file_path.suffix[1:] in _SUFFIXES_WITH_HEADER):
+            if not prog.search(file_path.read_text()):
+                report += f"{file_path.name}"
+    if report:
+        raise RuntimeError(f"One or more code files have missing or incorrect license header\n\n{report}")
```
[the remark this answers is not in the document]
JAlvarezJarreta: This is a mistake when I rewrote the function, I was aiming to do `report += f"{file_path.name}\n"`, if that's ok? To having to "go over the list more than once".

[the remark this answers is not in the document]
MatBarba: I usually prefer a list then join since it's "cleaner", but whichever is fine!

cicd/scripts/check_license.py
```diff
@@ -35,14 +36,16 @@
...
 
+    Args:
+        notice_template: Path to notice template file.
+
     Raises:
         RuntimeError: If the NOTICE file has incorrect format or the copyright year is not correct.
 
     """
-    notice_template = _TEMPLATES_DIR / "notice.tpl"
     notice_file = _ROOT_PATH / "NOTICE"
```
vsitnik, line 49: python is not for prepping :/

vsitnik (approved):

cicd/scripts/check_license.py
MatBarba: What do you mean by prepping?

vsitnik: I need to disable automatic spelling correction in mac os.
* grepping

MatBarba (approved):

PR #98, opened 2023-05-31
lcampbell/genome prepare

ens-LCampbell:
Added some slight changes to amend_genome_data and added option for collecting output files into a single user specified output directory. 

Note, all changes were done on top of @MatBarba most recent large scale code cleanup/refactoring commits to main branch.

### DIFF
--- a/pipelines/nextflow/subworkflows/genome_prepare/genome_prepare.nf
+++ b/pipelines/nextflow/subworkflows/genome_prepare/genome_prepare.nf
@@ -53,6 +53,9 @@ include { PROCESS_SEQ_REGION } from '../../modules/process_seq_region.nf'
 include { PROCESS_FASTA as PROCESS_FASTA_DNA } from '../../modules/process_fasta_data.nf'
 include { PROCESS_FASTA as PROCESS_FASTA_PEP } from '../../modules/process_fasta_data.nf'
 include { AMEND_GENOME_DATA } from '../../modules/amend_genome_data.nf'
+include { COLLECT_FILES; MANIFEST; PUBLISH_DIR } from '../../modules/collect_files.nf'
+include { CHECK_INTEGRITY } from '../../modules/integrity.nf'
+include { MANIFEST_STATS } from '../../modules/manifest_stats.nf'
 
 
 // Run main workflow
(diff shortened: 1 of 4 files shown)


src/python/ensembl/io/genomio/metadata/update_genome_metadata.py
```diff
@@ -183,7 +180,15 @@ class InputSchema(argschema.ArgSchema):
...
-    brc4_mode = argschema.fields.Int(required=False, metadata={"description": "Activate BRC4 mode (default)"})
+    brc4_mode = argschema.fields.Int(
+        required=False,
+        metadata={"description": "Activate BRC4 mode (default)"}
+    )
+    output_json = argschema.fields.OutputFile(
+        required=False,
+        dump_default="amend_genome_meta.json",
+        metadata={"description": "Default json file to capture json metadata"},
+    )
```
MatBarba, line 191: Minor: can you run black for this file?

pipelines/nextflow/subworkflows/genome_prepare/genome_prepare.nf
```diff
@@ -61,39 +64,47 @@ workflow GENOME_PREPARE {
...
+        //             .groupTuple()
+
+        // // Collect in manifest, checks and stats
+        // collect_dir = COLLECT_FILES(prepared_files)
+        // manifest_dired = MANIFEST(collect_dir, accession)
+        // manifest_checked = CHECK_INTEGRITY(manifest_dired, params.brc4_mode)
+        // manifest_stated = MANIFEST_STATS(manifest_checked, 'datasets')
+
+        // // Publish the data
+        // PUBLISH_DIR(manifest_stated, output_dir, accession)
```
MatBarba, line 109: I assume the pipeline is not finished at this point?

MatBarba (approved):
That looks fine to me

src/python/ensembl/io/genomio/metadata/update_genome_metadata.py
ens-LCampbell: Blacked ! Sorry my bad..... I had formatted it to be inline with other option params

pipelines/nextflow/subworkflows/genome_prepare/genome_prepare.nf
MatBarba: I think that's not going to work unless you keep the accession associated with each file along the whole pipeline (that's the only way to make sure we keep them in sync). This is really tricky and I'll need to focus on this once I'm done with other things...

That said, you can push the PR and we can work on that next part afterwards.

MatBarba (approved):

PR #99, opened 2023-05-31
Mbarba/cleanup3

MatBarba:
More work to make the code comply with pylint.
Several things remain to fix:
- integrity, manifest_stats, process_gff3 and assembly/get_assembly_data all have several functions that need to be refactored (Too many branches or  Too many nested blocks).
- The test files fail because pylint doesn't link us overloading pytest with our test file directories.
I will make separate PRs for each of those issues.
The present PR is heavy enough already to require a review before continuing to address the more specific issues.

### DIFF
--- a/src/python/ensembl/io/genomio/extract_from_gb.py
+++ b/src/python/ensembl/io/genomio/extract_from_gb.py
@@ -34,7 +34,7 @@ from collections import Counter
 import json
 from os import PathLike
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import argschema
 from Bio import SeqIO
@@ -207,8 +207,8 @@ class FormattedFilesGenerator:
 
     def _parse_record(self, record: SeqRecord) -> Tuple[SeqRecord, List[SeqRecord], List[str]]:
         all_ids: List[str] = []
-        peptides: List[SeqRecord] = []
-        feats = {}
+        peptides: List[SeqFeature] = []
+        feats: Dict[str, SeqFeature] = {}
 
         for feat in record.features:
             # Silently skip any unsupported feature type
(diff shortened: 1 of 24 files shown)


src/python/ensembl/io/genomio/extract_from_gb.py
```diff
@@ -26,33 +45,38 @@
...
-        "genome_data": formatter.genome,
-    }
+class GenomeFiles:
+    """Store the representation of the genome files created."""
+
+    def __init__(self, out_dir: Optional[PathLike] = None) -> None:
+        if out_dir is None:
+            out_dir = Path(".")
+        else:
+            out_dir = Path(out_dir)
```
JAlvarezJarreta, line 64: Alternative:
```suggestion
    def __init__(self, out_dir: PathLike = Path()) -> None:
```
From the documentation, `Path()` returns the current working directory.

src/python/ensembl/io/genomio/extract_from_gb.py
```diff
@@ -26,33 +45,38 @@
...
 
-    # Output the gff3 file
-    output = {
-        "gff3": formatter.genes_gff,
-        "fasta_dna": formatter.fasta_dna,
-        "fasta_pep": formatter.fasta_pep,
-        "seq_region": formatter.seq_region,
-        "genome_data": formatter.genome,
-    }
+class GenomeFiles:
```
JAlvarezJarreta, line 57: After checking the implementation and its usage, it may be worth considering as a future improvement making `class GenomeFiles(dict):`, so your `GenomeFiles` objects are dictionaries on their own (no need for the `to_dict` function).

JAlvarezJarreta (changes requested):
Nice work! Glad to see that there's clearly an improvement of the codebase after doing the exercise of listening to `pylint`. Some suggestions, and a couple of discussion comments (not with the intention to be done for this PR).

JAlvarezJarreta (changes requested):
Nice work! Glad to see that there's clearly an improvement of the codebase after doing the exercise of listening to `pylint`. Some suggestions, and a couple of discussion comments (not with the intention to be done for this PR).

src/python/ensembl/io/genomio/metadata/update_genome_metadata.py
```diff
@@ -143,20 +141,23 @@ def amend_genomic_metadata(
...
-
     # Get additional sequences in the assembly but not in the data
-    additions = get_additions(Path(INSDC_RefSeq_report_infile), Path(genbank_infile))
+    additions = get_additions(Path(insdc_refseq_report_infile), Path(genbank_infile))
     if additions:
         genome_metadata["added_seq"] = {"region_name": additions}
+
+    # Possible brc specific
+    if brc4_mode:
+        pass
```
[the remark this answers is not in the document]
MatBarba: yes

JAlvarezJarreta (approved):
Thank you for taking in most of my suggestions!

PR #110, opened 2023-06-14
Add coverage badge generation step to make it publicly available

JAlvarezJarreta:
I have also taken the opportunity to cleanup the debug job (to add a couple of variables and remove a duplicate) and updated the workflow and pages job rules.

_Note:_ the library I have added can allow us to create any additional badges we may need/want.


vsitnik (approved):
cool. thanks. though, could you add some comments on workflow/rules part? thanks

cicd/gitlab/dot.gitlab-ci.yml
```diff
@@ -17,9 +17,12 @@
 # https://docs.gitlab.com/ee/ci/yaml/workflow.html#switch-between-branch-pipelines-and-merge-request-pipelines
 workflow:
   rules:
-     - if: $CI_EXTERNAL_PULL_REQUEST_IID && $CI_COMMIT_BRANCH
-     - if: $CI_COMMIT_BRANCH =~ $CI_DEFAULT_BRANCH
-       when: never
+    - if: $CI_EXTERNAL_PULL_REQUEST_IID &&
+          $CI_EXTERNAL_PULL_REQUEST_SOURCE_REPOSITORY == $CI_EXTERNAL_PULL_REQUEST_TARGET_REPOSITORY &&
+          $CI_EXTERNAL_PULL_REQUEST_TARGET_BRANCH_NAME == $CI_DEFAULT_BRANCH
+    - if: $CI_COMMIT_BRANCH =~ $CI_DEFAULT_BRANCH
```
vsitnik, line 23: what does this mean?

JAlvarezJarreta: The intention with these rules is that only PRs originated within `ensembl-genomio` and against `main` will trigger an automatic run of the pipeline. For commit, only those done against `main` should trigger an automatic run of the pipeline, but not for any other (they can always be triggered manually from GitLab if needed).

vsitnik: Don't we have "when: never" on the following line?

cicd/gitlab/parts/python.gitlab-ci.yml
```diff
@@ -116,12 +116,15 @@ python:coverage:cobertura:
   extends: .python:coverage
   script:
     - coverage xml -o $RUN_DIR/cobertura_coverage.xml
+    - genbadge coverage -i $RUN_DIR/cobertura_coverage.xml -o $RUN_DIR/coverage-badge.svg
```
vsitnik, line 119: do we want to have coverages stages separated?
can we use `-o $RUN_DIR/coverage/coverage-badge.svg` so we won't need to `mv` anything in `pages` job?

JAlvarezJarreta: I have just realised `cobertura` is only useful for merge requests (GitLab), so it is useless for GitHub PRs. Will remove it and merge the coverage stages :+1:

cicd/gitlab/dot.gitlab-ci.yml
JAlvarezJarreta: Based on my interpretation of the documentation, if no rules apply, the pipeline is not created, e.g. https://docs.gitlab.com/ee/ci/yaml/workflow.html#workflow-rules-examples

PR #118, opened 2023-06-19
Process gff3 input only 1 tuple

MatBarba:
The previous version accepted 2 tuples, which led to desynchronisation.

Also using named output channels for readability.

### DIFF
--- a/pipelines/nextflow/subworkflows/genome_prepare/genome_prepare.nf
+++ b/pipelines/nextflow/subworkflows/genome_prepare/genome_prepare.nf
@@ -47,48 +47,54 @@ workflow GENOME_PREPARE {
     // Main data input to this subworkflow is genomic_dataset tuple
     main:        
         // Verify genome.json schema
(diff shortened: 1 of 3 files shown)


pipelines/nextflow/subworkflows/genome_prepare/genome_prepare.nf
```diff
@@ -47,48 +47,54 @@ workflow GENOME_PREPARE {
...
-                    CHECK_JSON_SCHEMA_SEQREG.out.verified_json,
-                    PROCESS_FASTA_DNA.out.processed_fasta)
+        prepared_files_grouped = amended_genome.concat(
+            gene_models, 
+            functional_annotation,
+            fasta_pep,
+            seq_region,
+            fasta_dna
+        )
+        prepared_files = prepared_files_grouped
```
ens-LCampbell, line 97: Thanks for cleaning up the code, much neater. Was on my todo list so thanks !

ens-LCampbell (approved):
Thanks for spotting the bug Mat. Mistake noted for future ref !

PR #123, opened 2023-06-28
modifying the parser to include tRNA and rRNA in gff

Dishalodha:
Adding support for tRNA and rRNA


JAlvarezJarreta (approved):
Approved without testing the code.

Dishalodha:
Note: we need to remove any special symbols from the ids

MatBarba (approved):
Approved with testing the code.

PR #126, opened 2023-07-11
Mbarba/fix nxf test

MatBarba:
Make various fixes to ensure the new Nextflow test works and behaves as expected.
- The manifest did not include the genome file, so I also fixed the pipeline for that part.
- The name of the test need to be simple ascii, since it is used in the path (it failed because of it for me).
- Also added requirements in pyproject.

NB: the Nextflow test does not yet run on CI/CD, but it should run locally with
`pytest pipelines/nextflow/tests/ --git-aware`
(git-aware to not copy the whole .git folder for the test)


src/python/ensembl/io/genomio/metadata/update_genome_metadata.py
```diff
@@ -123,32 +123,22 @@ def get_report_regions_names(report_path: Path) -> List[str]:
...
     insdc_refseq_report_infile: PathLike,
     genbank_infile: PathLike,
-    output_dir: PathLike,
     brc4_mode: Optional[int] = 1,
 ) -> Path:
     """
     Args:
         genome_infile: Genome data following the schemas/genome_schema.json.
+        genome_outfile: Amended genome data file.
+        output_dir: Directory where the new file will be created.
```
JAlvarezJarreta, line 135:
```suggestion
```

JAlvarezJarreta (approved):
I will read more about `pytest-workflow`, I'm curious. But everything looks good to me :+1:

PR #128, opened 2023-07-14
Use genomIO scripts

MatBarba:
In the loader pipeline


vsitnik:
testing using https://github.com/Ensembl/ensembl-production-metazoa/blob/main/meta/plants/110/rs8t_gsoja load

PR #129, opened 2023-07-14
Mbarba/loader fix fix

MatBarba:
Make pylint happy, but increase number of max attributes in objects to 12.


vsitnik (approved):
cool. thank you!

PR #133, opened 2023-07-18
Add ncbi_check option

MatBarba:
Only for assembly check,
so not used by sequence_addition

Set to 0 to skip the check

This is useful for e.g. the full pipeline run check (prevent failure, do not need to get data from NCBI, and ensure the stats file is not going to change).


vsitnik (approved):
ok to me

Dishalodha (approved):
Not tested but looks good to me

ens-LCampbell (approved):
Nice catch Matthieu ! I hadn't considered changing date implications on MD5 my bad sorry

PR #134, opened 2023-07-18
Compare gb id with both refseq and genbank

MatBarba:
Also remove version numbers for comparison

Currently the code would erronously list most of the IDs if from RefSeq.


vsitnik (approved):
used in amend_genomic_data (= "ensembl.io.genomio.metadata.update_genome_metadata:main")
ok for me

ens-LCampbell (approved):

Dishalodha (approved):

PR #135, opened 2023-07-18
Explicit stats_file path parameter

MatBarba:
The stats file should be wherever we want (and with the name we choose), not necessarily in the manifest directory. This is needed for better nextflow support.


vsitnik (approved):
ok for me

ens-LCampbell (approved):

Dishalodha (approved):

PR #136, opened 2023-07-19
Use genome/io check_json_schema

MatBarba:
No a complete transition, but enough to make the dumper pipeline run with the current code base


vsitnik (approved):
looks good to me

PR #137, opened 2023-07-21
Mbarba/better cache

MatBarba:
Using explicit file names, and paths different from the input in the output, allow Nextflow to cache those results properly.
Also set resources to 'adaptive' to a few more processes.

### DIFF
--- a/pipelines/nextflow/modules/gff3/process_gff3.nf
+++ b/pipelines/nextflow/modules/gff3/process_gff3.nf
@@ -17,19 +17,21 @@ nextflow.enable.dsl=2
 params.merge_split_genes="False"
 
 process PROCESS_GFF3 {
-    tag "$gff3 - $task.attempt"
+    tag "$gca"
     label 'variable_2_8_32'
-    errorStrategy 'ignore'
+    errorStrategy 'finish'
 
     input:
         tuple val(gca), path(gff3), path(genome)
 
     output:
-        tuple val(gca), path("*.json"), emit: functional_annotation
-        tuple val(gca), path("*.gff3"), emit: gene_models
+        tuple val(gca), path("functional_annotation.json"), emit: functional_annotation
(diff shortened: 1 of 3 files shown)


ens-LCampbell (approved):
Nothing I can see here, looks fine to me !

pipelines/nextflow/modules/gff3/process_gff3.nf
```diff
@@ -17,19 +17,21 @@ nextflow.enable.dsl=2
 params.merge_split_genes="False"
 
 process PROCESS_GFF3 {
-    tag "$gff3 - $task.attempt"
+    tag "$gca"
     label 'variable_2_8_32'
-    errorStrategy 'ignore'
+    errorStrategy 'finish'
```
Dishalodha, line 22: Does this mean, it will not proceed with other species but allow the already run jobs to finish in case of an error?

Dishalodha (approved):

pipelines/nextflow/modules/gff3/process_gff3.nf
MatBarba: Maybe I shouldn't include this here; but this means that it will allow the jobs already running to finish. Ignore would continue running all jobs, even the ones after that one (and we don't want that).

Dishalodha: so if a species enquires an error then the other species will just continue the job that is already running but not proceed further? this can lead to an incomplete run for other species? but alternatively can be resumes?

MatBarba: Yes, basically: let the pipeline finish what it's already started, fix the issue for the gff3 that failed, and resume

Dishalodha: okay seems good to me then, better to resume then missing the error

PR #138, opened 2023-07-21
Apply better resources for hungry processes

MatBarba:
Start adaptive at 4GB
Also include both 137 and 140 error codes for retry


Dishalodha (approved):
Looks good to me :)

PR #140, opened 2023-07-26
Mbarba/patch build

MatBarba:
Patch build related fixes:
- Allow selecting genes with creation date, for stable id allocation
- When running GFF3 loader batch pipeline, do not rename the existing logic_name, use the same one (if it has the same name). This is BRC specific.
- Add non_canonical split sites to skip in the process gff3 part
- Also, fix to use genomio AnalysisSetup instead of the protein-imported one


vsitnik (approved):
ok for me

PR #143, opened 2023-08-01
Mbarba/patch pipeline

MatBarba:
A whole new pipeline, but limited in scope: it's only to be used to update a core database after a patch build from VEuPathDB using the gene_diff script. It is using some older perl scripts, and some updated.

The pipeline goal is to do the following:
- Replace the stable ids by the old ones if the genes have been updated. Do the same for transcripts/translations if an updated gene still has some of the older transcripts.
- Transfer metadata from the previous database for those genes that have been updated, namely:
    - Description
    - Xrefs (excluding those from xref pipeline and protein features)
    - Version (incremented)
- Allocate stable ids to new genes and new transcripts using the OSID service.
- Load the events in the mapping_session and stable_id_event table (using the allocated IDs).
- Create standard exon stable ids and set the version for all features.

### DIFF
--- a/pipelines/nextflow/modules/patch_build/allocate_ids.nf
+++ b/pipelines/nextflow/modules/patch_build/allocate_ids.nf
@@ -32,7 +32,7 @@ process ALLOCATE_IDS {
     if (osid.mock) {
         osid_params = "--mock_osid"
     } else {
-        exit 1, "Actual OSID not implemented yet: $osid";
+        osid_params = "--osid_url $osid.url --osid_user $osid.user --osid_pass $osid.pass --organism $osid.species"
     }
     def feat_list = ""
     if (mode == "gene") {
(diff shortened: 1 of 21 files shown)


ens-LCampbell (approved):
I reviewed all 20 files commited. I haven't found any issues. Will there be example data to test the pipeline ? I guess testing would be tricky given it would also rely on new/old core_dbs ?

MatBarba:
> I reviewed all 20 files commited. I haven't found any issues. Will there be example data to test the pipeline ? I guess testing would be tricky given it would also rely on new/old core_dbs ?

Yeah, it's too complicated as a whole pipeline to test (and its use will be pretty limited so no point spending too much time here).

Best way to do this is to should always make a mock run first, check that the changes make sense, and then reload the db and run with the OSID service for the production run.

Dishalodha (approved):

PR #144, opened 2023-08-03
Mbarba/load gff apply patch

MatBarba:
Actually apply the translation seq edit patches (with a parameter on by default), like it is done in the genome loader pipeline.

Add a simple runnable to extract the db server parameters from the registry, so that no new parameters are needed to be given.

Also add a dummy LoadGFF3Start runnable for clarity.


ens-LCampbell (approved):

Dishalodha (approved):
Looks good to me :)

PR #148, opened 2023-08-17
Add Perl to our CI/CD

JAlvarezJarreta:
Although temporary, I believe it is still important to have `perlcritic` running as part of our CI/CD for the time being, since there is still debugging/small dev happening.

After running `perlcritic` I have tried to address all the issues raised. I believe none should produce a different outcome with the exception of the `eval` line change in `src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/AddSynonyms.pm`, since I'm not sure where the `{}` should be introduced (). Exact message error:
```
[BuiltinFunctions::ProhibitStringyEval] Expression form of "eval" at line 182, column 29.  (Severity: 5)
```
And its extended version:
```
The string form of `eval' is recompiled every time it is executed,
whereas the block form is only compiled once. Also, the string form
doesn't give compile-time warnings.

    eval "print $foo";        # not ok
    eval {print $foo};        # ok
```

### DIFF
--- a/src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/AnalysisSetup.pm
+++ b/src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/AnalysisSetup.pm
@@ -179,7 +179,7 @@ sub production_updates {
   ));
   $sth->fetch();
   
-  $properties{'web_data'} = eval ($properties{'web_data'}) if defined $properties{'web_data'};
+  $properties{'web_data'} = eval ({$properties{'web_data'}}) if defined $properties{'web_data'};
   
   # Explicitly passed parameters do not get overwritten.
   foreach my $property (keys %properties) {
--- a/src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/AnalysisSetup.pm
+++ b/src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/AnalysisSetup.pm
@@ -179,7 +179,7 @@ sub production_updates {
   ));
   $sth->fetch();
   
-  $properties{'web_data'} = eval {print $properties{'web_data'}} if defined $properties{'web_data'};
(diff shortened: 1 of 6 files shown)


src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/AnalysisSetup.pm
```diff
@@ -179,7 +179,7 @@ sub production_updates {
   ));
   $sth->fetch();
   
-  $properties{'web_data'} = eval ($properties{'web_data'}) if defined $properties{'web_data'};
+  $properties{'web_data'} = eval ({$properties{'web_data'}}) if defined $properties{'web_data'};
```
MatBarba, line 182:
```suggestion
  $properties{'web_data'} = eval "$properties{'web_data'}" if defined $properties{'web_data'};
```
This is a string eval, not a block (we're trying to load a data structure from a string here)

MatBarba:
```suggestion
  $properties{'web_data'} = eval {print $properties{'web_data'}} if defined $properties{'web_data'};
```
Rather this

src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/AnalysisSetup.pm
```diff
@@ -179,7 +179,7 @@ sub production_updates {
   ));
   $sth->fetch();
   
-  $properties{'web_data'} = eval ($properties{'web_data'}) if defined $properties{'web_data'};
+  $properties{'web_data'} = eval {print $properties{'web_data'}} if defined $properties{'web_data'};
```
JAlvarezJarreta, line 182:
```suggestion
  $properties{'web_data'} = eval { $properties{'web_data'} } if defined $properties{'web_data'};
```

PR #149, opened 2023-08-18
Add data files to GenomIO Python library

JAlvarezJarreta:
I know we are not fans of symlinks and configuration files outside `pyproject.toml`, but this is the easiest way (and I have been testing options for more than 2 hours) to incorporate the data files into our Python library.

This is the new standard procedure to fetch and use data files in Python: https://docs.python.org/3/library/importlib.resources.html
_Note:_ Since most functionalities have been added in 3.9, the temporary dependency I have included allows to produce code almost 100% compatible in 3.8+.

`MANIFEST.in` is a requirement of setuptools and it is not currently supported as part of `pyproject.toml` ([decision under review](https://github.com/pypa/setuptools/issues/3341)). Poetry has already added support for this, so it may be another reason so **add** support for it in the near future.


Dishalodha (approved):
Looks good to me, not sure but will it be useful to add `nextflow.md` file in Manifest.in?

JAlvarezJarreta:
> Looks good to me, not sure but will it be useful to add `nextflow.md` file in Manifest.in?

I'm happy to do so, but to my understanding if a user wants the pipelines they will download the whole repository, not just do `pip install ensembl-genomio` (notice here I'm referring to a pypy instance, not a path).

ens-LCampbell (approved):
I need to read up a bit more but from what you've added Im happy to approve the addition

PR #150, opened 2023-08-24
Apply updates to SLURM profile following conversation with TSC

JAlvarezJarreta:
https://www.nextflow.io/docs/latest/config.html?highlight=queueglobalstatus#scope-executor

Taken the opportunity to simplify the profiles putting the variables in their corresponding blocks to ease learning and tracking. I have also merged other suggestions from the EBI Nextflow config proposal [[link](https://github.com/nf-core/configs/blob/master/conf/ebi_codon.config)].

### DIFF
--- a/pipelines/nextflow/workflows/nextflow.config
+++ b/pipelines/nextflow/workflows/nextflow.config
@@ -70,6 +70,7 @@ profiles {
         process {
             executor = 'lsf'
             queue = { task.memory < 300.GB ? 'standard' : 'bigmem' }
+            queueGlobalStatus = true
             perJobMemLimit = true
             errorStrategy = 'finish'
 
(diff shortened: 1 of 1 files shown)


pipelines/nextflow/workflows/nextflow.config
```diff
@@ -29,7 +32,7 @@ profiles {
     slurm {
         process {
             executor = 'slurm'
-            queue = { task.memory < 300.GB ? 'standard' : 'bigmem' }
+            queueGlobalStatus = true
```
ens-LCampbell, line 35: Haha this param seems rather vital 👍

ens-LCampbell (approved):

Dishalodha (approved):
Nice! looks good to me

JAlvarezJarreta:
> I can see no issues with these updates. Why can I ask do we explicitly list these exit codes ? 104, 137, 140, 143 ?

We had 140 and TSC has put in their Nextflow + Slurm document the other 3, so I have gone for a merge rather than following their suggestion because 140 has worked well for us so far.

PR #152, opened 2023-09-19
Include transcript ids to allocate

MatBarba:
Include a useful intermediate file, containing the IDs of the new transcripts for genes where the stable ID is conserved.


JAlvarezJarreta (approved):
Approved in the software development meeting.

PR #155, opened 2023-09-26
Adjust GitLab CI/CD based on documentation

JAlvarezJarreta:
The file changes suggests to not use `only` anymore as it is not under active development, and use `rules` instead. Furthermore, I've seen a few examples where root files are surrounded by `"`, so I've done so for `pyproject.toml`, just in case.


JAlvarezJarreta:
It seems the suggested changes are not compatible with our current set-up. Since it seems GitLab CI is running as expected now, will drop this for the time being.

PR #157, opened 2023-10-05
Group species params into a meta key

MatBarba:
Cleaner approach: all species specific information are put in the meta key.

### DIFF
--- a/pipelines/nextflow/workflows/additional_seq_prepare/main.nf
+++ b/pipelines/nextflow/workflows/additional_seq_prepare/main.nf
@@ -50,15 +50,20 @@ if (params.help) {
     exit 0
 }
 
-assert params.accession, "Parameter 'acession' is not specified"
+assert params.accession, "Parameter 'accession' is not specified"
 assert params.prefix, "Parameter 'prefix' is not specified"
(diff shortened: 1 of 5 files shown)


pipelines/nextflow/workflows/additional_seq_prepare/main.nf
```diff
@@ -50,15 +50,20 @@ if (params.help) {
     exit 0
 }
 
-assert params.accession, "Parameter 'acession' is not specified"
+assert params.accession, "Parameter 'accession' is not specified"
 assert params.prefix, "Parameter 'prefix' is not specified"
```
vsitnik, line 54: Are we allowed to use an empty string? Should we use empty by default?

vsitnik (approved):

pipelines/nextflow/workflows/additional_seq_prepare/main.nf
vsitnik: update pipelines/nextflow/modules/genbank/extract_from_gb.nf  to use `""` for prefix.
check  that extract_from_gb  is allowed to have an empty suffix

PR #162, opened 2023-10-05
Collect files is used by other pipelines, fallback

MatBarba:
Fix the broken genome_prepare pipeline (introduced by the partial meta key update in my previous PR).


vsitnik (approved):
cool. thanks!

JAlvarezJarreta (approved):

PR #163, opened 2023-10-05
Minor folder structure changes

JAlvarezJarreta:
I have also taken the opportunity to update the `subworkflows/yaml-schema.json` to match the one in nf-core: https://github.com/nf-core/modules/blob/master/subworkflows/yaml-schema.json


MatBarba (approved):
Looks ok

PR #168, opened 2023-10-10
argparser equivalent to current argschema

JAlvarezJarreta:
This is an example of an `argparse` code equivalent to `argschema`, going to the extent of having the same input file check functionality and also making the type being `pathlib.Path` instead of string (which is not a bad additional feature).

Whilst a functionality to check (and if not create) output directories makes sense, that is not the case for input files/directories: the Python's mantra of "ask for forgiveness not for permission" states that you should assume the input parameters are correct and handle the exception "beautifully" if that is not the case. Thus, checking input files/directories exists in the parsing step may be a functionality we may not want to pursue anymore. The only part that we definitely need to sit down and evaluate is when we have taken advantage of `--output_json` from `argschema`.

### DIFF
--- a/src/python/ensembl/io/genomio/metadata/prepare_genome.py
+++ b/src/python/ensembl/io/genomio/metadata/prepare_genome.py
@@ -27,6 +27,7 @@ __all__ = [
     "DEFAULT_API_URL",
 ]
 
+import argparse
 import datetime
 from os import PathLike
 from pathlib import Path
@@ -34,7 +35,6 @@ from typing import Dict, Optional
 from xml.etree import ElementTree
 from xml.etree.ElementTree import Element
 
-import argschema
 import requests
 
 from ensembl.io.genomio.utils import get_json, print_json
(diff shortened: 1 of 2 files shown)


src/python/ensembl/io/genomio/metadata/prepare_genome.py
```diff
@@ -255,22 +258,31 @@ def prepare_genome_metadata(
...
-    mod = argschema.ArgSchemaParser(schema_type=InputSchema)
-    prepare_genome_metadata(mod.args["json_file"], mod.args["output_dir"])
+    parser = argparse.ArgumentParser(
+        description=("Expand the genome metadata with information about the provider, assembly and gene"
+                     "build version, and taxonomy."),
+    )
+    parser.add_argument(
+        "--json_file",
+        required=True,
+        type=lambda x: validate_input_path(parser, x),
```
MatBarba, line 277: Would this still work if the parameter was not required and no file was provided?

src/python/ensembl/io/genomio/metadata/prepare_genome.py
```diff
@@ -255,22 +258,31 @@ def prepare_genome_metadata(
...
+        help="Genome metadata JSON file path",
+    )
+    parser.add_argument(
+        "--output_dir",
+        default=Path().absolute(),
+        type=Path,
+        help="Output folder for the updated genome metadata JSON file (default: current working directory)",
```
MatBarba, line 284: We can show the default automatically with an argument to the parser (https://stackoverflow.com/questions/12151306/argparse-way-to-include-default-values-in-help):
```
parser = argparse.ArgumentParser(
    # ... other options ...
    formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```

PR #173, opened 2023-10-25
Added new features following B67 production

JAlvarezJarreta:
Added new include/exclude features, including some additional processing for IG/T-cell elements to be converted to the corresponding biotype.

We also agreed that it seems the best that the code merges split genes by default and generates a report of the merged genes that can be reviewed in case something doesn't look right.

### DIFF
--- a/src/python/ensembl/brc4/runnable/fill_metadata.py
+++ b/src/python/ensembl/brc4/runnable/fill_metadata.py
@@ -116,14 +116,13 @@ class fill_metadata(eHive.BaseRunnable):
 
     def run_sql_req(self, sql, log_pfx, from_file=False):
         os.makedirs(dirname(log_pfx), exist_ok=True)
-        en_root = self.param_required("ensembl_root_dir")
 
         sql_option = r""" -sql '{_sql}' """.format(_sql=sql)
         if from_file:
             sql_option = r""" < '{_sql}' """.format(_sql=sql)
 
         cmd = r"""{_dbcmd} -url "{_srv}{_dbname}" {_sql_option} > {_out} 2> {_err}""".format(
-            _dbcmd="perl %s/ensembl-hive/scripts/db_cmd.pl" % (en_root),
+            _dbcmd="perl %s/scripts/db_cmd.pl" % os.getenv("EHIVE_ROOT_DIR"),
             _srv=self.param("dbsrv_url"),
             _dbname=self.param("db_name"),
             _sql_option=sql_option,
--- a/src/python/ensembl/io/genomio/gff3/process_gff3.py
+++ b/src/python/ensembl/io/genomio/gff3/process_gff3.py
@@ -925,10 +925,10 @@ class InputSchema(argschema.ArgSchema):
         default=True, metadata={"description": "Generate and add stable IDs when missing?"}
     )
     out_gff_path = argschema.fields.OutputFile(
-        dump_default="gene_models.gff3", metadata={"description": "Output gff path"}
+        default="gene_models.gff3", metadata={"description": "Output gff path"}
     )
     out_func_path = argschema.fields.OutputFile(
-        dump_default="functional_annotation.json",
+        default="functional_annotation.json",
         metadata={"description": "Output functional_annotation.json path"},
     )
     retain_split_genes = argschema.fields.Boolean(
--- a/src/python/ensembl/io/genomio/gff3/process_gff3.py
+++ b/src/python/ensembl/io/genomio/gff3/process_gff3.py
@@ -342,7 +342,7 @@ class GFFSimplifier(GFFParserCommon):
             transcript: Gene segment transcript feature.
 
         """
-        # Change mobile_genetic_element into a transposable_element feature
+        # Change V/C_gene_segment into a its corresponding transcript names
         if transcript.type in ("C_gene_segment", "V_gene_segment"):
             standard_name = transcript.qualifiers["standard_name"][0]
             # Drop "_segment" from the transcript type
@@ -922,7 +922,7 @@ class InputSchema(argschema.ArgSchema):
     in_gff_path = argschema.fields.InputFile(required=True, metadata={"description": "Input gene.gff3 path"})
     genome_data = argschema.fields.InputFile(metadata={"description": "genome.json path"})
     make_missing_stable_ids = argschema.fields.Boolean(
-        default=True, metadata={"description": "Generate and add stable IDs when missing?"}
+        default=True, metadata={"description": "Generate and add stable IDs when missing or invalid?"}
     )
     out_gff_path = argschema.fields.OutputFile(
         default="gene_models.gff3", metadata={"description": "Output gff path"}
--- a/src/python/ensembl/io/genomio/gff3/process_gff3.py
+++ b/src/python/ensembl/io/genomio/gff3/process_gff3.py
@@ -130,7 +130,7 @@ class GFFParserCommon:
 class GFFGeneMerger(GFFParserCommon):
     """Specialized class to merge split genes in a GFF3 file, prior to further parsing."""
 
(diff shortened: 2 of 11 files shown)


src/python/ensembl/io/genomio/integrity.py
```diff
@@ -733,10 +732,7 @@ def main() -> None:
     """Main entrypoint."""
     mod = argschema.ArgSchemaParser(schema_type=InputSchema)
 
-    # Start
-    inspector = IntegrityTool(mod.args["manifest_file"])
-    if mod.args.get("brc_mode"):
-        inspector.set_brc_mode(True)
+    inspector = IntegrityTool(mod.args["manifest_file"], mod.args["brc_mode"])
```
MatBarba, line 735: If br_mode is not provided in the args, this crashes with KeyError. Fix:
```suggestion
    inspector = IntegrityTool(mod.args["manifest_file"], mod.args.get("brc_mode", False))
```

MatBarba (changes requested):

src/python/ensembl/io/genomio/integrity.py
JAlvarezJarreta: This is one of those changes that I missed. The idea is to set the default value of `brc_mode` to false, so it acts as a flag. Will apply it now.

src/python/ensembl/io/genomio/gff3/process_gff3.py
```diff
@@ -325,6 +333,32 @@ def format_mobile_element(self, feat):
...
+    def format_gene_segments(self, transcript: SeqFeature) -> SeqFeature:
+        """Returns the equivalent Ensembl biotype feature for gene segment transcript features.
+
+        Supported features: "C_gene_segment" and "V_gene_segment".
+
+        Args:
+            transcript: Gene segment transcript feature.
+
+        """
+        # Change mobile_genetic_element into a transposable_element feature
```
Dishalodha, line 345:
```suggestion
        # Change V/C_gene_segment into a its corresponding transcript names
```

Dishalodha (commented):
Do we want to make merge_split_genes as default before having checks ? I am afraid that we might miss some gff issues which do not fall into the category of merge split genes but we might not want to keep this flag as default until we have better checks. 
Following this conversation on slack [https://genomes-ebi.slack.com/archives/C03GP4JKMFS/p1696847910854649]. 
However otherwise we could add 'retain_split_genes' parameter in our init command everytime until we have checks

JAlvarezJarreta:
> Do we want to make merge_split_genes as default before having checks ? I am afraid that we might miss some gff issues which do not fall into the category of merge split genes but we might not want to keep this flag as default until we have better checks. Following this conversation on slack [https://genomes-ebi.slack.com/archives/C03GP4JKMFS/p1696847910854649]. However otherwise we could add 'retain_split_genes' parameter in our init command everytime until we have checks

I think I misunderstood the conversation, because what you are saying makes sense. Happy to apply the change you suggest.

src/python/ensembl/io/genomio/gff3/process_gff3.py
```diff
@@ -325,6 +333,32 @@ def format_mobile_element(self, feat):
...
+
+        Args:
+            transcript: Gene segment transcript feature.
+
+        """
+        # Change V/C_gene_segment into a its corresponding transcript names
+        if transcript.type in ("C_gene_segment", "V_gene_segment"):
+            standard_name = transcript.qualifiers["standard_name"][0]
+            # Drop "_segment" from the transcript type
+            biotype = transcript.type[:-8]
```
MatBarba, line 349: One line with a more explicit code:
```suggestion
            biotype = transcript.type.replace("_segment", "")
```

src/python/ensembl/io/genomio/gff3/process_gff3.py
```diff
@@ -325,6 +333,32 @@ def format_mobile_element(self, feat):
...
+            if re.search(r"\b(immunoglobulin|ig)\b", standard_name, flags=re.IGNORECASE):
+                biotype = f'IG_{biotype}'
+            elif re.search(r"\bt[- _]cell\b", standard_name, flags=re.IGNORECASE):
+                biotype = f'TR_{biotype}'
+            else:
+                print(f"Unexpected 'standard_name' content for feature {transcript.id}: {standard_name}")
```
MatBarba, line 355: We might want to keep this logged instead of just printing, or even exit since we need to check how to handle that

src/python/ensembl/io/genomio/gff3/process_gff3.py
```diff
@@ -414,6 +448,9 @@ def _normalize_transcripts(self, gene: SeqFeature, fail_types) -> SeqFeature:
             transcript_number = count + 1
             transcript.id = self.normalize_transcript_id(gene.id, transcript_number)
 
+            if transcript.type in ("C_gene_segment", "V_gene_segment"):
```
MatBarba, line 451: This check is done twice: here, and in the format_gene_segments. Might as well keep only the one in the method, but that'll require removing the "else" condition in it.

src/python/ensembl/io/genomio/integrity.py
MatBarba, line 735: Better be explicit in case we want to make changes later:
```suggestion
    inspector = IntegrityTool(mod.args["manifest_file"], brc_mode=mod.args["brc_mode"])
```

MatBarba (approved):
Some suggestions, but nothing major

src/python/ensembl/brc4/runnable/fill_metadata.py
```diff
@@ -116,14 +116,13 @@ def flattern(self, data, ignore_lst, pfx=None):
...
         os.makedirs(dirname(log_pfx), exist_ok=True)
-        en_root = self.param_required("ensembl_root_dir")
 
         sql_option = r""" -sql '{_sql}' """.format(_sql=sql)
         if from_file:
             sql_option = r""" < '{_sql}' """.format(_sql=sql)
 
         cmd = r"""{_dbcmd} -url "{_srv}{_dbname}" {_sql_option} > {_out} 2> {_err}""".format(
-            _dbcmd="perl %s/ensembl-hive/scripts/db_cmd.pl" % (en_root),
+            _dbcmd="perl %s/scripts/db_cmd.pl" % os.getenv("EHIVE_ROOT_DIR"),
```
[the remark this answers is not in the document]
JAlvarezJarreta: The codebase definitely does: https://github.com/search?q=repo%3AEnsembl%2Fensembl-hive+EHIVE_ROOT_DIR&type=code

src/python/ensembl/io/genomio/gff3/process_gff3.py
```diff
@@ -941,12 +938,11 @@ def main() -> None:
...
-    num_merged_genes = merger.merge(in_gff_path, interim_gff_path)
-
-    # If there are split genes, decide to merge, or just die
+    merged_genes = merger.merge(in_gff_path, interim_gff_path)
+    num_merged_genes = len(merged_genes)
     if num_merged_genes > 0:
-        if mod.args["retain_split_genes"]:
-            raise GFFParserError("GFF contains split genes. Fix it or remove '--retain_split_genes' flag.")
+        # Report the list of merged genes in case something does not look right
+        print(f"{num_merged_genes} genes merged:\n" + "\n".join(merged_genes))
```
Dishalodha, line 945: Should we output a log file with the GCA to store the merge gene event?

JAlvarezJarreta: At the moment we have no logging system in place, so a comprehensive print is the best we can do. Since this is stored in Nextflow's log file, and it is one per job, I guess it is ok. But feel free to correct me if I'm wrong.

PR #174, opened 2023-10-25
Updates in the Genome prepare pipeline following B67 production

JAlvarezJarreta:
Several code fixes and updates to make the pipeline work as expected. These fixes were run during B67 production, so it has been tested.


pipelines/nextflow/modules/genome_metadata/dump_ncbi_stats.nf
```diff
@@ -16,13 +16,12 @@
 
 process DUMP_NCBI_STATS {
     tag "${db.species}"
-    label "local"
-    storeDir "$cache_dir/ncbi_stats/${db.species}"
+    label 'local'
+    local 'cached'
```
vsitnik, line 20: label??

pipelines/nextflow/subworkflows/genome_prepare/main.nf
```diff
@@ -35,71 +34,64 @@ include { CHECK_INTEGRITY } from '../../modules/manifest/integrity.nf'
...
         ncbi_check
 
-    // Main data input to this subworkflow is genomic_dataset tuple
     main:
-        // We expect every input and output stream to have `meta` as the first val in the form of:
-        //   tuple("accession": accession, "production_name": production_name, "prefix": prefix)
-
         // Verify genome.json schema
-        checked_genome = CHECK_JSON_SCHEMA_GENOME(genomic_dataset)
+        checked_genome = CHECK_JSON_SCHEMA_GENOME(genomic_dataset).verified_json
```
vsitnik, line 47: do we need `.verified_json`? it's the only bit that's emitted.

pipelines/nextflow/subworkflows/genome_prepare/main.nf
```diff
@@ -35,71 +34,64 @@ include { CHECK_INTEGRITY } from '../../modules/manifest/integrity.nf'
...
 
-        // Download genome data files. Files may or may not include gene models (gff3) and/or peptides.
-        (download_min, download_opt) = DOWNLOAD_ASM_DATA(checked_genome, cache_dir)
+        // Download genome data files. Files may or may not include gene models (GFF3) and/or peptides.
+        DOWNLOAD_ASM_DATA(checked_genome)
+        download_min = DOWNLOAD_ASM_DATA.out.min_set
+        download_opt = DOWNLOAD_ASM_DATA.out.opt_set
 
-        // Check for presence of annotation related output files, then process gene models data if present
-        if (download_opt) {
```
vsitnik, line 60: why are we removing `if (dwonload_opt)`?

vsitnik (commented):
cool. thanks. just a couple of questions.

pipelines/nextflow/modules/genome_metadata/dump_ncbi_stats.nf
JAlvarezJarreta: Yep, missed that!

pipelines/nextflow/subworkflows/genome_prepare/main.nf
JAlvarezJarreta: Because the logic of those assemblies with annotation is handled by the channel creation itself (`genome_gff_files = checked_genome.join(unpacked_gff, failOnDuplicate: true)`).

pipelines/nextflow/modules/gff3/process_gff3.nf
```diff
@@ -20,7 +20,7 @@ process PROCESS_GFF3 {
     label 'cached'
 
     input:
-        tuple val(meta), path(gff3), path(genome)
+        tuple val(meta), path(genome), path(gff3)
```
vsitnik, line 23: I hope, this is supported both in `genome_prepare` and `additional_seq_prepare subworkflows`. I see the patch. Just confirming.

vsitnik (approved):
cool. thanks!

pipelines/nextflow/modules/gff3/process_gff3.nf
```diff
@@ -13,23 +13,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-params.merge_split_genes="False"
 
 process PROCESS_GFF3 {
     tag "${meta.accession}"
     label 'adaptive'
+    label 'cached'
```
MatBarba, line 20:
```suggestion
```
This should not be cached: if we make changes to the process (e.g. adding acceptable biotypes), then we need to have it rerun, but without redownloading the data from genbank).

pipelines/nextflow/modules/gff3/process_gff3.nf
JAlvarezJarreta: Yes, I have tried to be thorough and any changes in the `input` or `output` are applied to every pipeline affected.

pipelines/nextflow/workflows/nextflow.config
```diff
@@ -88,4 +95,8 @@ process {
         errorStrategy = { task.exitStatus in [104, 137, 140, 143] ? 'retry' : 'finish' }
         maxRetries = 3
     }
+    withLabel: 'cached' {
```
MatBarba, line 98: This needs to be added to the tests nextflow config as well, otherwise the pipelines tests are broken since they can't use the cached data that we've prepared.

Alternatively, put this into a common nextflow file, and load it in both this file and the test file.

pipelines/nextflow/modules/manifest/manifest_maker.nf
```diff
@@ -14,8 +14,9 @@
 // limitations under the License.
 
 process MANIFEST {
-    tag "Manifest_$meta.accession"
+    tag "manifest_$meta.accession"
     label 'adaptive'
+    label 'cached'
```
MatBarba, line 19: This should not be cached as well: this is data that we generate and we don't reuse it.

MatBarba (changes requested):
The current version breaks the pipeline tests (mostly because of how the cache handling works).

A few tweaks are necessary to make it work.

The addition prepare pipeline also requires an update, but I can't make a suggestion in unchanged files (cf separate PR #175).

pipelines/nextflow/workflows/nextflow.config
MatBarba: NB: I believe this is the main remaining blocking part for the pipeline tests to run correctly

JAlvarezJarreta: I wonder if you can "redefine" labels, i.e. you can have a `nextflow.config` in the tests area that loads this one first, and then changes the content of those labels again here (so for instance things like `retry_on_error` (if introduced) and `cached` could stay as they are).

pipelines/nextflow/subworkflows/genome_prepare/main.nf
vsitnik: ok. thanks

pipelines/nextflow/subworkflows/genome_prepare/main.nf
vsitnik: ok. cool. thanks!

MatBarba (approved):
Everything is fine now :)

ens-LCampbell (approved):

PR #176, opened 2023-10-25
Mbarba/fix genome prep

MatBarba:
Grouped my suggestions: those should be enough to make the pipelines tests work.


pipelines/nextflow/tests/workflows/nextflow_test.config
```diff
@@ -40,6 +40,7 @@ process {
 
     withLabel: 'local' {
         executor = 'local'
+        maxRetries = 1
```
JAlvarezJarreta, line 43: By default `maxRetries` should be one, so I believe you can skip this line.

JAlvarezJarreta (approved):

PR #172, opened 2023-10-24
Address some missing elements to control module imports correctly

JAlvarezJarreta:
I haven't done this for the single-file modules, as I believe some of these may be relocated soon.

PR #178, opened 2023-10-27
Bugfixes and code updated on GenomIO utils

JAlvarezJarreta:
- ".gz" can also be found in ".tar.gz" for instance, so the latter needs to be checked always first
- `Path.suffix` would only grab the last ".*", ignoring cases such as ".tar.gz"
- `shutil.unpack_archive()` creates the destination if it does not exist, so this behaviour has been replicated in the other cases for consistency

### DIFF
--- a/src/python/ensembl/io/genomio/utils/archive_utils.py
+++ b/src/python/ensembl/io/genomio/utils/archive_utils.py
@@ -61,6 +61,7 @@ def extract_file(src_file: PathLike, dst_dir: PathLike) -> None:
 
     """
     src_file = Path(src_file)
+    # Create a set of all the possible "right-side suffixes", e.g. from "myfile.tar.gz" you get {".tar.gz", ".gz"}
     extensions = {"".join(src_file.suffixes[i:]) for i in range(0, len(src_file.suffixes))}
     dst_dir = Path(dst_dir)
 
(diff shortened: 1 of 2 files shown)


ens-LCampbell (approved):

src/python/ensembl/io/genomio/utils/archive_utils.py
```diff
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Utils to deal with archived files (gzip)."""
+"""Utils to deal with archive files."""
 
 __all__ = ["SUPPORTED_ARCHIVE_FORMATS", "extract_file"]
```
MatBarba, line 17:
```suggestion
__all__ = ["SUPPORTED_ARCHIVE_FORMATS", "extract_file", "open_gz_file"]
```

MatBarba (approved):
Not a fan of all these abbreviations, but the code looks sound.

src/python/ensembl/io/genomio/utils/archive_utils.py
JAlvarezJarreta: This change has been addressed in another PR (https://github.com/Ensembl/ensembl-genomio/pull/172), thus why I haven't changed it.

PR #179, opened 2023-10-30
Addition of workflow config for sphinx doc generation

ens-LCampbell:
Contents of PR:
- single YAML file to set up config of workflow. This workflow is restricted to trigger only when a PR is made to main. PR includes main as accepted branch, and excludes any branch linked to current Ensmetazoa team members. 
- Slight modification to sphinx build script (addition of new script called in context of workflow.)
- Workflow includes params set to reuse python installation deps with -cache to speed up re workflow after initial run.

### DIFF
--- a/.github/workflows/sphinx_doc_generation.yaml
+++ b/.github/workflows/sphinx_doc_generation.yaml
@@ -1,0 +1,49 @@
+name: Documentation_deploy
+run-name: ${{ github.actor }} triggered doc generation
+on: 
+  pull_request:
+    branches:
+      - main
+      - '!jalvarez/**'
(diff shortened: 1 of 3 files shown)


MatBarba (approved):
Looks great! One thing missing I guess are the scripts themselves (and their list of params from argparse/Argschema), but I'm not sure how we could get them here.

MatBarba:
One note: the Usage shows this:
```
3 pip install -r requirements.txt -e .
```

In our case to install we use pyproject instead of requirements.txt, so:

`pip install -e .`

Not sure where that's defined

JAlvarezJarreta (approved):

PR #196, opened 2023-11-07
Bugfix: include initial event for each release

JAlvarezJarreta:
Moreover, I have made the class' messages more self-explanatory.


MatBarba (approved):
That's a bug introduced in March (2023-03-31), so I'll hunt the core dbs impacted from there (might be a dozen of so)

PR #197, opened 2023-11-07
Make a explicit CI/CD tag

JAlvarezJarreta:
This is the best course of action if we want to ensure only the bare minimum tools used in CI/CD are included (but no other libraries that may ease development work). If we want to support a `dev` tag as well that holds `cicd` and `doc` together without explicitly duplicating everything, the best option at the moment is to have a script that can automatically create a `requirements-dev.txt` file every time `pyproject.toml` is updated (so it can be installed with `pip install -r requirements-dev.txt`).

_Note 1:_ not sure if needed, not sure if useful, but worth flagging this up.
_Note 2:_ I have left `ensembl-py` where it is since this is being addressed in another PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Though, this bit is in constant development. Some documentatin can be found in [
 See [docs](docs)
 
 ## Unit testing
-The Python part of the codebase has now unit tests available to test each module. Make sure you have installed this repository's `[dev]` dependencies (via `pip install ensembl-genomio[dev]`) before continuing.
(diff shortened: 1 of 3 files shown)


vsitnik (approved):
Cool. Thanks. But we need to update
https://github.com/Ensembl/ensembl-genomio#optional-installation

vsitnik (commented):
https://github.com/Ensembl/ensembl-genomio/blob/a6f627da64c0c9c1bb8840b73815619f60c5d008/README.md#unit-testing   as well

Dishalodha (approved):

ens-LCampbell (approved):

MatBarba (approved):
Should be good after Vasily's suggestions are applied

vsitnik (approved):
thanks!

PR #198, opened 2023-11-08
Additional B67 updates

JAlvarezJarreta:
The main change is to avoid generating an empty GFF3 file early in the pipeline if there are no records to avoid `gt` to fail later on (cannot handle empty GFF3 files). I have opted for this approach because it makes more sense to not flow anything when there is nothing to process (saving compute) than handling empty files, i.e. it requires more manual intervention to asses if the GFF3 was supposed to be empty at that point.

### DIFF
--- a/pipelines/nextflow/modules/files/publish_output.nf
+++ b/pipelines/nextflow/modules/files/publish_output.nf
@@ -14,9 +14,9 @@
 // limitations under the License.
 
 process PUBLISH_DIR {
-    publishDir "$out_dir/${meta.accession}", mode: 'copy', overwrite: false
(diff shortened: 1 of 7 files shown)


MatBarba (approved):
Pipeline test pass, but the current test uses a genbank file that has annotations (which should be the most common case).

This brings up a point about missing annotations: if an additional sequence doesn't have annotations, we need to have a warning at the end so that we can check with outreach to make sure this is expected.

This might be resolved by the logging system, or we'll need to think about a report stage at the end.

Dishalodha (approved):
Looks good

PR #205, opened 2023-11-09
Adding log to seq_region module

Dishalodha:
Seq_region module

1. Dump.py: did not have a lot of print messages so just removed some commented lines
2. Prepare.py: I have used warn where necessary, I have not used the debug option but I have mentioned it in the args just incase we use it in future.

### DIFF
--- a/src/python/ensembl/io/genomio/seq_region/dump.py
+++ b/src/python/ensembl/io/genomio/seq_region/dump.py
@@ -252,4 +252,3 @@ def main() -> None:
         seq_regions = get_seq_regions(session, external_map)
 
     print(json.dumps(seq_regions, indent=2))
-
--- a/src/python/ensembl/io/genomio/seq_region/dump.py
+++ b/src/python/ensembl/io/genomio/seq_region/dump.py
@@ -24,6 +24,7 @@ from sqlalchemy.orm import Session, joinedload
 from ensembl.core.models import CoordSystem, SeqRegion, SeqRegionSynonym, SeqRegionAttrib
 from ensembl.database import DBConnection
 from ensembl.utils.argparse import ArgumentParser
+from ensembl.utils.logging import init_logging_with_args
 
 
 ROOT_DIR = Path(__file__).parent / "../../../../../.."
@@ -238,7 +239,9 @@ def main() -> None:
     parser.add_argument_src_path(
         "--external_db_map", default=DEFAULT_MAP.resolve(), help="File with external_db mapping"
     )
+    parser.add_log_arguments()
     args = parser.parse_args()
+    init_logging_with_args(args)
 
     dbc = DBConnection(args.url)
 
--- a/src/python/ensembl/io/genomio/seq_region/dump.py
+++ b/src/python/ensembl/io/genomio/seq_region/dump.py
@@ -80,7 +80,6 @@ def get_seq_regions(session: Session, external_db_map: dict) -> List[SeqRegion]:
     seq_regions = []
 
     for coord_system in coord_systems:
-        # print(f"Dump coord {coord_system.name}")
         seqr_stmt = (
             select(SeqRegion)
             .where(SeqRegion.coord_system_id == coord_system.coord_system_id)
@@ -251,6 +250,3 @@ def main() -> None:
 
     print(json.dumps(seq_regions, indent=2))
 
-
-if __name__ == "__main__":
-    main()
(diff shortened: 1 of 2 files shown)


src/python/ensembl/io/genomio/seq_region/prepare.py
```diff
@@ -439,6 +442,8 @@ def prepare_seq_region_metadata(
     gbff_file: Optional[PathLike] = None,
     brc_mode: bool = False,
     to_exclude: Optional[List[str]] = None,
+    debug:  bool = False,
+    verbose:  bool = False,
```
JAlvarezJarreta, line 446:
```suggestion
    debug: bool = False,
    verbose: bool = False,
```

src/python/ensembl/io/genomio/seq_region/prepare.py
```diff
@@ -498,6 +503,19 @@ def main() -> None:
...
         "--to_exclude", nargs="*", metavar="SEQ_REGION_NAME", help="Sequence region names to exclude"
     )
+    parser.add_argument("-d", "--debug", action="store_true", help="Debug level logging")
+    parser.add_argument("-v", "--verbose", action="store_true", help="Verbose level logging")
+
     args = parser.parse_args()
+    
+    log_format = "%(asctime)s\t%(levelname)s\t%(message)s"
+    date_fmt = r"%Y-%m-%d %H:%M:%S"
+    reload(logging)
```
JAlvarezJarreta, line 513: What is this aiming for?

src/python/ensembl/io/genomio/seq_region/prepare.py
```diff
@@ -498,6 +503,19 @@ def main() -> None:
...
+    if args.verbose:
+        logging.basicConfig(format=log_format, datefmt=date_fmt, level=logging.INFO)
+    elif args.debug:
+        logging.basicConfig(format=log_format, datefmt=date_fmt, level=logging.DEBUG)
+    else:
+        logging.basicConfig(format=log_format, datefmt=date_fmt, level=logging.WARNING)
 
-    prepare_seq_region_metadata(**vars(args))
+    prepare_seq_region_metadata(**vars(args))
```
MatBarba, line 521: I think it's better to give the explicit list and handle the verbose/debug in the main, rather than pass them to prepare_seq_region_metadata()

src/python/ensembl/io/genomio/seq_region/prepare.py
```diff
@@ -498,6 +503,19 @@ def main() -> None:
...
+    
+    log_format = "%(asctime)s\t%(levelname)s\t%(message)s"
+    date_fmt = r"%Y-%m-%d %H:%M:%S"
+    reload(logging)
+    if args.verbose:
+        logging.basicConfig(format=log_format, datefmt=date_fmt, level=logging.INFO)
+    elif args.debug:
+        logging.basicConfig(format=log_format, datefmt=date_fmt, level=logging.DEBUG)
+    else:
+        logging.basicConfig(format=log_format, datefmt=date_fmt, level=logging.WARNING)
```
MatBarba, line 519: Should be rewritten to use the new ensembl-py argparse and init_logging

MatBarba (changes requested):
Need to be rewritten to handle the new argparse/init_logging system

Also need to handle the verbose/debug options in the main, so do not pass all params to prepare_seq_region_metadata()

MatBarba (approved):
I have added the standard add_log_arguments() and init_logging_with_args() to both scripts

src/python/ensembl/io/genomio/seq_region/dump.py
```diff
@@ -239,7 +239,9 @@ def main() -> None:
     parser.add_argument_src_path(
         "--external_db_map", default=DEFAULT_MAP.resolve(), help="File with external_db mapping"
     )
+    parser.add_log_arguments()
     args = parser.parse_args()
+    init_logging_with_args(args)
```
JAlvarezJarreta, line 244: Do we want to make this the standard, even if there is only one log argument involved?

JAlvarezJarreta (approved):

src/python/ensembl/io/genomio/seq_region/dump.py
MatBarba: Yeah, no need to think, less risk of error if we want to add a log_file

PR #207, opened 2023-11-09
Example of how to initialise the new logging framework

JAlvarezJarreta:
This is the example on how the new `ensembl.utils.logging` module (https://github.com/Ensembl/ensembl-py/pull/35) should be applied to our codebase.

Example of help output:
```bash
usage: assembly_download [-h] [--download_dir PATH] [-v | --debug | --log LEVEL] [--log_file PATH] [--log_file_level LEVEL] accession

Download an assembly data files from INSDC or RefSeq.

positional arguments:
  accession             Genome assembly accession

optional arguments:
  -h, --help            show this help message and exit
  --download_dir PATH   Folder where the data will be downloaded (default: /hps/software/users/ensembl/repositories/metazoa/jalvarez/dev/ensembl-py)

logging arguments:
  -v, --verbose         Verbose mode, i.e. 'INFO' log level
  --debug               Debugging mode, i.e. 'DEBUG' log level
  --log LEVEL           Level or severity of the events to track: DEBUG, INFO, WARNING, ERROR, CRITICAL (default: WARNING)
  --log_file PATH       Log file path (default: None)
  --log_file_level LEVEL
                        Level or severity of the events to track in the log file: DEBUG, INFO, WARNING, ERROR, CRITICAL (default: DEBUG)
```


src/python/ensembl/io/genomio/assembly/download.py
```diff
@@ -281,6 +282,8 @@ def main() -> None:
     parser.add_argument_dst_path(
         "--download_dir", default=Path.cwd(), help="Folder where the data will be downloaded"
     )
+    parser.add_log_arguments()
```
ens-LCampbell, line 285: Not quite sure where add_log_arguments is inherited from

MatBarba: I assume from a future update to ensembl-py's ArgumentParser (so we can't use that PR right now)

ens-LCampbell: That was my guess yep

JAlvarezJarreta: That's correct. I do not want to do any PRs into `ensembl-py` until we like what we have here.

vsitnik:
what's the use case? where's is logging is called from? I can see only initalisation.

MatBarba:
> How is it used? where's is logging is called from? I can see only initialisation :( I'm definitely missing something.

There's a logging.info() in the rest of the code already (not part of the PR)

MatBarba (changes requested):
Looks ok, though will need to remove the logging init done in retrieve_assembly_data(()

JAlvarezJarreta:
> Looks ok, though will need to remove the logging init done in retrieve_assembly_data(()

I believe all the remaining PRs will need to be changed if this is merged. Fortunately, the change should be minimal and most of it easily done with a `sed` or similar.

src/python/ensembl/io/genomio/_logging.py
```diff
@@ -0,0 +1,45 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Internal logging system for the Ensembl GenomIO modules."""
+
+import logging
+from pathlib import Path
+from typing import Optional
+
+
+def init_logging(log_level: str = "WARNING", log_file: Optional[Path] = None) -> None:
```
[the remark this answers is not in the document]
MatBarba: OK, that should work, I didn't realize setLevel accepted strings.

Another point then that I missed, I would definetely/might use init_logging() with logging.WARNING, which is an int, so should this accept Union[str, int] for log_level?

src/python/ensembl/io/genomio/_logging.py
```diff
@@ -0,0 +1,45 @@
...
+        log_level: Minimum logging level for the standard error.
+        log_file: Logging file where to write debug (and above) logging messages.
+
+    """
+    formatter = logging.Formatter("%(asctime)s\t%(levelname)s\t%(message)s", datefmt=r"%Y-%m-%d_%H:%M:%S")
+    # Set the standard error logging handler and replace the existing one in the root logging system
+    stream_handler = logging.StreamHandler()
+    stream_handler.setLevel(log_level)
+    stream_handler.setFormatter(formatter)
+    logging.root.handlers[0] = stream_handler
```
MatBarba, line 38: That's basically what [basicConfig](https://docs.python.org/3/library/logging.html#logging.basicConfig) does, so:
```suggestion
    logging.basicConfig(format="%(asctime)s\t%(levelname)s\t%(message)s", datefmt=r"%Y-%m-%d_%H:%M:%S", level=log_level)
```

src/python/ensembl/io/genomio/_logging.py
```diff
@@ -0,0 +1,45 @@
...
+    if log_file:
+        # Set the debug-level logging file handler and add it to the root logging system
+        logfile_handler = logging.FileHandler(log_file)
+        logfile_handler.setLevel(logging.DEBUG)
+        logfile_handler.setFormatter(formatter)
+        logging.root.addHandler(logfile_handler)
+        logging.root.setLevel(logging.DEBUG)
```
MatBarba, line 45: Already defined with basicConfig
```suggestion
```

src/python/ensembl/io/genomio/_logging.py
JAlvarezJarreta: Great suggestion :+1: Applied in the `ensembl-py` PR.

ens-LCampbell (approved):
Looks really good ! I think this setup should suffice most peoples logging needs. I think once we're happy get this merged to allow some further testing 👍

PR #208, opened 2023-11-10
Fix to genome_stats compare + json utils

ens-LCampbell:
Noticed that genome_stats.compare was failing when operating over organ_abbrevs with empty ncbi_stats JSON files.

- Implemented logging in genome_stats.compare 
- Added try block to get_json, returns an empty json instead of just failing

### DIFF
--- a/src/python/ensembl/io/genomio/genome_stats/compare.py
+++ b/src/python/ensembl/io/genomio/genome_stats/compare.py
@@ -191,15 +191,13 @@ def main() -> None:
     )
     parser.add_argument_src_path("--ncbi_stats", required=True, help="NCBI dataset JSON file")
     parser.add_argument_src_path("--core_stats", required=True, help="Core database JSON file")
-    parser.add_argument("-v", "--verbose", action="store_true", required=False,
-        help="Verbose level logging")
-    parser.add_argument("-d", "--debug", action="store_true", required=False,
-        help="Debug level logging")
+    parser.add_argument("-v", "--verbose", action="store_true", required=False, help="Verbose level logging")
+    parser.add_argument("-d", "--debug", action="store_true", required=False, help="Debug level logging")
     args = parser.parse_args()
 
     # Configure logging
-    date_format='%Y/%m/%d_%I:%M:%S(%p)'
-    logging_format='%(asctime)s - %(levelname)s - %(message)s'
+    date_format = "%Y/%m/%d_%I:%M:%S(%p)"
+    logging_format = "%(asctime)s - %(levelname)s - %(message)s"
     reload(logging)
     log_level = None
     if args.debug:
@@ -207,16 +205,13 @@ def main() -> None:
     elif args.verbose:
         log_level = logging.INFO
 
-    logging.basicConfig(
-        format=logging_format, datefmt=date_format,
-        filemode="w", level=log_level
-        )
+    logging.basicConfig(format=logging_format, datefmt=date_format, filemode="w", level=log_level)
 
     try:
         ncbi_stats = get_json(args.ncbi_stats)["reports"][0]
         logging.info(f"{args.ncbi_stats} JSON .'reports' obtained")
-    except KeyError:
-        logging.warning(f"{args.ncbi_stats} JSON is empty")
+    except (json.decoder.JSONDecodeError, KeyError):
+        logging.warning(f"{args.ncbi_stats} JSON file is empty")
         ncbi_stats = {}
     core_stats = get_json(args.core_stats)
     all_stats = compare_stats(ncbi_stats, core_stats)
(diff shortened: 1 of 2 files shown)


src/python/ensembl/io/genomio/genome_stats/compare.py
```diff
@@ -189,11 +191,32 @@ def main() -> None:
...
+    parser.add_argument("-v", "--verbose", action="store_true", required=False,
+        help="Verbose level logging")
+    parser.add_argument("-d", "--debug", action="store_true", required=False,
+        help="Debug level logging")
     args = parser.parse_args()
 
+    # Configure logging
+    date_format='%Y/%m/%d_%I:%M:%S(%p)'
+    logging_format='%(asctime)s - %(levelname)s - %(message)s'
+    reload(logging)
```
JAlvarezJarreta, line 203: I'm still trying to figure out why this is needed (and, hence, if it is needed in my PR). I've seen this in @Dishalodha's PR as well, so I have removed the "if" and left only the "why" question.

src/python/ensembl/io/genomio/utils/json_utils.py
```diff
@@ -30,7 +29,11 @@ def get_json(src_path: PathLike, **kwargs) -> Any:
 
     """
     with Path(src_path).open("r") as json_file:
-        return json.load(json_file, **kwargs)
+        try:
+            return json.load(json_file, **kwargs)
+        # Empty JSON file
+        except:
+            return {}
```
JAlvarezJarreta, line 36: Do we really want to return an empty dictionary when the JSON file cannot be read? As a general approach, I believe raising the exception is a better option, and leaving the `try`-`except` for the function calling `get_json()`.

JAlvarezJarreta: I'm afraid I would need to know more to be able to reply exactly. But if we are talking about cases like [`genome_stats/compare.py`](https://github.com/Ensembl/ensembl-genomio/blob/hackathon_nov23/src/python/ensembl/io/genomio/genome_stats/compare.py#L194-L197) where the code is currently only handling the file existing but being empty, I would then add `FileNotFoundError` to avoid stopping the pipeline, i.e. `except (FileNotFoundError, KeyError):`.

src/python/ensembl/io/genomio/genome_stats/compare.py
```diff
@@ -189,11 +191,32 @@ def main() -> None:
...
+        log_level = logging.INFO
+
+    logging.basicConfig(
+        format=logging_format, datefmt=date_format,
+        filemode="w", level=log_level
+        )
+
     try:
         ncbi_stats = get_json(args.ncbi_stats)["reports"][0]
+        logging.info(f"{args.ncbi_stats} JSON .'reports' obtained")
```
JAlvarezJarreta, line 217: This logging message should be in the `else` clause (in case you are not familiar with this, see second code block above [here](https://docs.python.org/3/tutorial/errors.html#raising-exceptions)).

src/python/ensembl/io/genomio/genome_stats/compare.py
```diff
@@ -189,11 +191,32 @@ def main() -> None:
...
     except KeyError:
+        logging.warning(f"{args.ncbi_stats} JSON is empty")
```
JAlvarezJarreta, line 219:
```suggestion
        logging.warning(f"{args.ncbi_stats} JSON file is empty")
```

src/python/ensembl/io/genomio/genome_stats/compare.py
```diff
@@ -189,11 +191,32 @@ def main() -> None:
...
+    log_level = None
+    if args.debug:
+        log_level = logging.DEBUG
+    elif args.verbose:
+        log_level = logging.INFO
+
+    logging.basicConfig(
+        format=logging_format, datefmt=date_format,
+        filemode="w", level=log_level
+        )
```
JAlvarezJarreta, line 213: Remember to run `black` over your changes (I have found myself many times with little things I missed).

src/python/ensembl/io/genomio/utils/json_utils.py
JAlvarezJarreta: In the case of an empty file, I would add `json.decoder.JSONDecodeError` to the `except` block instead.

src/python/ensembl/io/genomio/genome_stats/compare.py
ens-LCampbell: Black ran, sorry I always forget this step !

src/python/ensembl/io/genomio/utils/json_utils.py
ens-LCampbell: I've amended the Raise Error as suggested, revery json utils and ran black on code changes. Tested also in full nxflow pipeline run.

PR #213, opened 2023-11-15
Mbarba/genbank logging2

MatBarba:
Implement logging with `ensembl.utils.argparse` and `ensembl.utils.logging` from `ensembl-py` for the 2 genbank modules

### DIFF
--- a/src/python/ensembl/io/genomio/genbank/extract_data.py
+++ b/src/python/ensembl/io/genomio/genbank/extract_data.py
@@ -178,7 +178,7 @@ class FormattedFilesGenerator:
             all_ids += rec_ids
             peptides += rec_peptides
 
-        logging.debug(f"Write {len(records)} gff records to {self.files['gene_models']}")
+        logging.debug(f"Write {len(records)} gene records to {self.files['gene_models']}")
         if records:
             with self.files["gene_models"].open("w") as gff_fh:
                 GFF.write(records, gff_fh)
@@ -366,9 +366,8 @@ class FormattedFilesGenerator:
             if codon_table is None:
                 logging.warning(
                     (
-                        "No codon table found. "
-                        f"Make sure to change the codon table number in {self.files['seq_region']} manually "
-                        "if it is not the standard codon table"
+                        "No codon table found. Make sure to change the codon table number in "
+                        f"{self.files['seq_region']} manually if it is not the standard codon table."
                     )
                 )
 
@@ -387,9 +386,8 @@ class FormattedFilesGenerator:
                 if not codon_table:
                     logging.warning(
                         (
-                            f"'{seq.organelle}' is an organelle: "
-                            f"make sure to change the codon table number in {self.files['seq_region']} "
-                            "manually if it is not the standard codon table"
+                            f"'{seq.organelle}' is an organelle: make sure to change the codon table number "
+                            f"in {self.files['seq_region']} manually if it is not the standard codon table"
                         )
                     )
 
--- a/src/python/ensembl/io/genomio/genbank/extract_data.py
+++ b/src/python/ensembl/io/genomio/genbank/extract_data.py
@@ -483,13 +483,16 @@ def main() -> None:
     parser.add_argument("-d", "--debug", action="store_true", help="Debug level logging")
     args = parser.parse_args()
 
+    # Logging setup
+    log_level = None
+    if args.debug:
+        log_level = logging.DEBUG
+    elif args.verbose:
+        log_level = logging.INFO
     logging_format = "%(asctime)s\t%(levelname)s\t%(message)s"
     date_format = r"%Y-%m-%d_%H:%M:%S"
     reload(logging)
-    if args.verbose:
-        logging.basicConfig(format=logging_format, datefmt=date_format, level=logging.INFO)
-    if args.debug:
-        logging.basicConfig(format=logging_format, datefmt=date_format, level=logging.DEBUG)
+    logging.basicConfig(format=logging_format, datefmt=date_format, level=log_level)
 
     gb_extractor = FormattedFilesGenerator(prefix=args.prefix, prod_name=args.prod_name, gb_file=args.gb_file)
     gb_extractor.extract_gb(args.out_dir)
--- a/src/python/ensembl/io/genomio/genbank/extract_data.py
+++ b/src/python/ensembl/io/genomio/genbank/extract_data.py
@@ -43,11 +43,7 @@ from Bio.SeqRecord import SeqRecord
 from Bio.SeqFeature import SeqFeature
 
 from ensembl.utils.argparse import ArgumentParser
-
-
-logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -174,23 +178,23 @@ def _write_genes_gff(self) -> None:
             all_ids += rec_ids
             peptides += rec_peptides
 
-        # Write the annotated records to a clean GFF
+        logging.debug(f"Write {len(records)} gff records to {self.files['gene_models']}")
```
JAlvarezJarreta, line 181:
```suggestion
        logging.debug(f"Write {len(records)} gene records to {self.files['gene_models']}")
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -347,10 +351,10 @@ def _uniquify_id(self, gene_id, all_ids):
         new_id = gene_id
         num = 1
         while new_id in all_ids:
-            print(f"{new_id} exists, update")
+            logging.info(f"{new_id} already exists, append a number to the new one to make it unique")
```
JAlvarezJarreta, line 354:
```suggestion
            logging.info(f"{new_id} already exists: appending a number to the new one to make it unique")
```
And I would move this message before the loop, as I believe notifying once and then later what number it has been should be enough.

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -360,9 +364,9 @@ def _write_seq_region_json(self):
         for seq in self.seq_records:
             codon_table = self._get_codon_table(seq)
             if codon_table is None:
-                print(
+                logging.warning(
                     (
-                        "Warning: No codon table found. "
+                        "No codon table found. "
                         f"Make sure to change the codon table number in {self.files['seq_region']} manually "
                         "if it is not the standard codon table"
```
JAlvarezJarreta, line 371:
```suggestion
                        "No codon table found. Make sure to change the codon table number in "
                        f"{self.files['seq_region']} manually if it is not the standard codon table."
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -381,9 +385,9 @@ def _write_seq_region_json(self):
             if seq.organelle:
                 seq_obj["location"] = self._prepare_location(seq.organelle)
                 if not codon_table:
-                    print(
+                    logging.warning(
                         (
-                            f"Warning: '{seq.organelle}' is an organelle: "
+                            f"'{seq.organelle}' is an organelle: "
                             f"make sure to change the codon table number in {self.files['seq_region']} "
                             "manually if it is not the standard codon table"
```
JAlvarezJarreta, line 392:
```suggestion
                            f"'{seq.organelle}' is an organelle: make sure to change the codon table number "
                            f"in {self.files['seq_region']} manually if it is not the standard codon table"
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -398,16 +402,16 @@ def _write_seq_region_json(self):
                 },
             }
             if not seq_obj["added_sequence"]["assembly_provider"]["name"]:
-                print(
+                logging.warning(
                     (
-                        "Warning: please add the relevant provider name"
+                        "Please add the relevant provider name"
                         f"for the assembly in {self.files['seq_region']}"
                     )
```
JAlvarezJarreta, line 409:
```suggestion
                    f"Please add the relevant provider name for the assembly in {self.files['seq_region']}"
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -398,16 +402,16 @@ def _write_seq_region_json(self):
...
                 )
             if not seq_obj["added_sequence"]["assembly_provider"]["url"]:
-                print(
+                logging.warning(
                     (
-                        "Warning: please add the relevant provider url"
+                        "Please add the relevant provider url"
                         f" for the assembly in {self.files['seq_region']}"
                     )
```
JAlvarezJarreta, line 416:
```suggestion
                    f"Please add the relevant provider URL for the assembly in {self.files['seq_region']}"
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -455,8 +459,8 @@ def _write_genome_json(self):
         }
 
         if not genome_data["species"]["production_name"]:
-            print(
-                f"Warning: please add the relevant production_name for this genome in {self.files['genome']}"
+            logging.warning(
+                f"Please add the relevant production_name for this genome in {self.files['genome']}"
```
JAlvarezJarreta, line 463:
```suggestion
                f"Please add the relevant production name for this genome in {self.files['genome']}"
```
Just to follow the same format as previous messages (unless this is a slightly different case).

JAlvarezJarreta (changes requested):
A few minor edits, but looks good overall. Glad to see adding a logging has resulted in such a simple set of changes.

src/python/ensembl/io/genomio/genbank/extract_data.py
MatBarba: That suggestion can't be commited because it contains a deleted line (the one with "Warning:")

src/python/ensembl/io/genomio/genbank/extract_data.py
MatBarba: I would keep production_name because it's a specific term that we use (not just a name for production).

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -469,13 +465,15 @@ def _write_genome_json(self):
...
+    parser.add_argument_src_path("--gb_file", required=True, help="sequence accession file")
+    parser.add_argument("--prefix", required=True, help="prefix to add to every feature ID")
+    parser.add_argument("--prod_name", required=True, help="production name for the species")
     parser.add_argument_dst_path(
-        "--out_dir", default=Path.cwd(), help="Output folder where the generated files will be stored"
+        "--out_dir", default=Path.cwd(), help="output folder where the generated files will be stored"
     )
+    parser.add_log_arguments()
     args = parser.parse_args()
+    init_logging(args.log_level)
```
vsitnik, line 476: do we need to pass `args.log_file`, `args.log_file_level` as well? Thanks

vsitnik (approved):
thanks

src/python/ensembl/io/genomio/genbank/extract_data.py
JAlvarezJarreta: Those arguments are not available with `parser.add_log_arguments()` (without setting the flag of this method to `True`).

JAlvarezJarreta (approved):

PR #214, opened 2023-11-15
Implement ensembl-py logging 

ens-LCampbell:
Implement use of ens-py logging for ensembl.io. modules:

genomio.assembly.download
genomio.events.dump
genomio.events.load
genomio.genome_stats.compare
genomio.manifest.check_integrity
genomio.metadata.dump

- Module updated to use ensembl-py.utilis.logging
- Nextflow modules set to use --verbose as default


src/python/ensembl/io/genomio/genome_stats/compare.py
```diff
@@ -22,10 +22,11 @@
 import json
 import re
 from typing import Any, Dict
+import logging
 
 from ensembl.io.genomio.utils import get_json
 from ensembl.utils.argparse import ArgumentParser
-
+from ensembl.utils.logging import init_logging
```
JAlvarezJarreta, line 29:
```suggestion
from ensembl.utils.logging import init_logging

```

src/python/ensembl/io/genomio/genome_stats/compare.py
```diff
@@ -189,17 +190,22 @@ def main() -> None:
...
+    # Configure and initialise logging
+    init_logging(args.log_level, args.log_file, args.log_file_level)
+    
     try:
         ncbi_stats = get_json(args.ncbi_stats)["reports"][0]
-    except KeyError:
+        logging.info(f"{args.ncbi_stats} JSON .'reports' obtained")
+    except (json.decoder.JSONDecodeError, KeyError):
+        logging.warning(f"{args.ncbi_stats} JSON file is empty")
         ncbi_stats = {}
```
JAlvarezJarreta, line 204:
```suggestion
    except (json.decoder.JSONDecodeError, KeyError):
        logging.warning(f"{args.ncbi_stats} JSON file is empty")
        ncbi_stats = {}
    else:
        logging.info(f"{args.ncbi_stats} JSON .'reports' obtained")
```
As a note, the info message does not look too explanatory for me, but it may be because I'm not too familiar with what the code does.

src/python/ensembl/io/genomio/genome_stats/compare.py
```diff
@@ -189,17 +190,22 @@ def main() -> None:
...
     core_stats = get_json(args.core_stats)
     all_stats = compare_stats(ncbi_stats, core_stats)
 
     print(json.dumps(all_stats, indent=2, sort_keys=True))
 
-
```
JAlvarezJarreta, line 203: I believe this line should stay based on my recollection of PEP8 rules.

JAlvarezJarreta (changes requested):

MatBarba (changes requested):
Nice overall, though there should be no need to catch exceptions just to log them

src/python/ensembl/io/genomio/assembly/download.py
```diff
@@ -281,6 +281,10 @@ def main() -> None:
...
         "--download_dir", default=Path.cwd(), help="Folder where the data will be downloaded"
     )
+    parser.add_log_arguments(add_log_file=True)
     args = parser.parse_args()
 
-    retrieve_assembly_data(**vars(args))
+    # Configure and initialise logging
+    init_logging(args.log_level, args.log_file, args.log_file_level)
+
+    retrieve_assembly_data(args.accession, args.download_dir)
```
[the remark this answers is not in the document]
ens-LCampbell: I understand its a simple task, but it can not hurt having it in place. If you don't need a log file, don't call it with that args.param I guess :)

src/python/ensembl/brc4/runnable/core_server.py
```diff
@@ -55,7 +56,10 @@ def connect(self) -> None:
         )
 
     def set_database(self, db_name: str) -> None:
-        self._connector.database = db_name
+        try:
+            self._connector.database = db_name
+        except:
+            logging.critical(f"Unknown database: '{db_name}' - Not located on host !")
```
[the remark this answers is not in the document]
ens-LCampbell: Thanks @MatBarba. I've updated the way this is handled, can you take a look when you have time thanks

docs/gen_ref_pages.py
```diff
@@ -4,7 +4,9 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
+
 # http://www.apache.org/licenses/LICENSE-2.0
+
```
JAlvarezJarreta, line 9:
```suggestion
#
# http://www.apache.org/licenses/LICENSE-2.0
#
```
Not 100% sure, but there is a chance this might trigger an error in the license header check.

pipelines/nextflow/modules/download/download_asm_data.nf
```diff
@@ -36,6 +36,6 @@ process DOWNLOAD_ASM_DATA {
 
     shell:
         '''
-        assembly_download --accession !{meta.accession} --asm_download_dir ./
+        assembly_download --accession !{meta.accession} --download_dir ./ --verbose
```
JAlvarezJarreta, line 39: I have a feeling we chatted about this but cannot remember the conclusion: do we want to run always our code with `--verbose` by default? If so it "better" to convert this flag to `--quiet` and set `logging.INFO` as the default level, rather than `logging.WARNING`.

src/python/ensembl/brc4/runnable/core_server.py
```diff
@@ -55,8 +57,12 @@ def connect(self) -> None:
         )
 
     def set_database(self, db_name: str) -> None:
-        self._connector.database = db_name
-
+        try:
+            self._connector.database = db_name
+        except:
+            logging.exception(f"Unknown database ! DB:'{db_name}' not located on host {self.host}")
+            raise OperationalError(f"Unknown database!", None, None)
```
JAlvarezJarreta, line 64: It is advisable to only handle the exception you are interested in, and I would also suggest to just re-raise after the logging step:
```suggestion
        except mysql.connector.errors.ProgrammingError:
            logging.exception(f"Unknown database ! DB:'{db_name}' not located on host {self.host}")
            raise
```

src/python/ensembl/brc4/runnable/core_server.py
```diff
@@ -17,8 +17,10 @@
 import argparse
 import re
 from typing import Dict, List, Any
+import logging
 
 import mysql.connector
+from  sqlalchemy.exc import OperationalError
```
JAlvarezJarreta, line 23: If following my next comment, this can go:
```suggestion
```

JAlvarezJarreta (changes requested):
Looks good to me. It has also helped raise a couple of questions regarding how we want to set the default system + Exception handling that should be discussed in the next sofware development meeting.

JAlvarezJarreta (approved):

MatBarba (approved):

PR #220, opened 2023-11-16
Enable trace file by default with useful field information

JAlvarezJarreta:
With the inclusion of a logging system, this information is not stored in `.nextflow.log` but in each process' `.command.log` file (for the "default" StreamHandler, i.e. stderr). Thus I believe we need to enable the trace by default (will generate a `trace-<timestamp>.txt` by default where `nextflow run` is being executed) so we have an easy way to access a given process' working directory (and log file).

I have also left a comment with guidelines in how to use filters, in case we believe we want to only generate traces for failed tasks and/or specific modules for some pipelines.

There is also a full detail HTML report option ([nextflow run ... -with-report](https://www.nextflow.io/docs/latest/tracing.html#execution-report)) but I believe this may be a little bit overkilling in most cases (still useful to analyse a production run and see if we can refine some resources allocations).

### DIFF
--- a/pipelines/nextflow/workflows/nextflow.config
+++ b/pipelines/nextflow/workflows/nextflow.config
@@ -25,6 +25,12 @@ params {
     brc_mode = 0
 }
 
+trace {
+    enabled = true
+    fields = 'task_id,name,process,hash,status,exit,attempt,submit,time,realtime,cpus,%cpu,memory,%mem'
+//    filter = 'name =~ /foo.*/ && status == "FAILED"'
+}
+
 profiles {
 
     standard {


ens-LCampbell (changes requested):
This addition is really useful and welcome IMO. Would be nice to also include at least a first draft template to format the report if we want to use `nextflow log -t`. There are plenty of fields we can use to customise it https://www.nextflow.io/docs/latest/tracing.html#trace-fields

pipelines/nextflow/README.md
```diff
@@ -7,6 +7,16 @@ Please, refer to the "Installation" section of the [Nextflow pipelines document]
...
 
+### Trace and log reports
+Add `-with-trace` to your `nextflow run ...` command to activate the generation of a trace file with useful information such as resource usage of each job and their hash directory. More information can be found in the [Nextflow documentation](https://www.nextflow.io/docs/latest/tracing.html).
+
+Most of our jobs in our pipelines have the `--verbose` flag set as default, providing a more detailed log. You can generate a single log report with the information from every job by running:
+```bash
+nextflow log <run name> -t log_template.md > log_report.md
+```
+
+However, sometimes this report may be too large to be useful, so the previous command includes a `-filter` option, e.g. `-filter 'name =~ /DOWNLOAD_ASM_DATA.*/'` will provide a log report only for those jobs whose name starts with `DOWNLOAD_ASM_DATA`.
```
MatBarba, line 18: A note about this: "[...] this report may be too large to be useful [...]"

This is inevitable if we process a large number of species, but if we look at each genome, we need the INFO logs to be at least readable: i.e. what we write to INFO needs to be relatively short and informative, so should ideally be less than a couple of pages long per process.

(That said if the logs are full of warnings we do need all the details)

MatBarba (approved):

ens-LCampbell (approved):
Looks good to me. However, what is the rational for having two NXF README docs on in pipelines/nextflow and one in docs/nextflow ? Do they need to be separate files or could we combined all into the docs/nextflow.md ?

JAlvarezJarreta:
> Looks good to me. However, what is the rational for having two NXF README docs on in pipelines/nextflow and one in docs/nextflow ? Do they need to be separate files or could we combined all into the docs/nextflow.md ?

It was already there :see_no_evil: But yes, it makes sense to reduce documentation where possible to avoid duplication of efforts. Will create a JIRA ticket for that since it is out of the scope of this PR.

PR #226, opened 2023-11-24
Mbarba/logging standard

MatBarba:
Add add_log_arguments() + init_logging_with_args() to all scripts to make things standard.

Also includes some fixes to the test pipelines.


pipelines/nextflow/tests/workflows/test_genome_prepare_brc_mode_on.yml
```diff
@@ -48,3 +48,8 @@
       md5sum: 585da9f97f094e83702860ce43da652f
     - path: ./test_genome_prepare_output/GCA_017607445.1/stats.txt
       md5sum: 6104869d437ec4a9d13a2d070c307b0f
+  stdout:
+    contains:
+      - "N E X T F L O W"
```
JAlvarezJarreta, line 53: This looks annoying :see_no_evil:

src/python/ensembl/io/genomio/assembly/download.py
```diff
@@ -281,10 +281,8 @@ def main() -> None:
     parser.add_argument_dst_path(
         "--download_dir", default=Path.cwd(), help="Folder where the data will be downloaded"
     )
-    parser.add_log_arguments(add_log_file=True)
+    parser.add_log_arguments()
```
JAlvarezJarreta, line 284: I believe Vasily mentioned if we wanted to have the `add_log_file=True` set by default in our codebase just to have the option if needed/wanted.

JAlvarezJarreta (approved):
Looks good to me. Remember to update the corresponding JIRA tickets after it has been merged.

src/python/ensembl/io/genomio/assembly/download.py
MatBarba: I removed that one without thinking because it's just a (fast) download so I don't see a use for a log file. I'd just add that option if we actually need to be a detailled introspection at some point.

Though if we do decide to set this as default I'll follow of course

pipelines/nextflow/tests/workflows/test_genome_prepare_brc_mode_on.yml
MatBarba: This might be overdoing it since nextflow either worked or not... I think I'll remove the stdout checks

src/python/ensembl/io/genomio/assembly/download.py
JAlvarezJarreta: I prefer to add it as needed, to be honest. So let's move on.

ens-LCampbell (approved):
Looks good thanks Matthieu

PR #227, opened 2023-11-27
Move JSON schemas as part of the data files of the Python library

JAlvarezJarreta:
For now I have opted for a "fixed" approach, i.e. schemas can only be passed by name, matching one of the existing schema files, on the CLI. This is because some additional changes may be required to add the option for `--json_schema_file` parameter since it would need to be put into a mutually exclusion group in the parser, and then the `add_argument_src_path()` method would not be available. There are ways to replicate the behaviour, but I do not believe it is worth it unless you tell me otherwise.

_Note:_ I have not checked if the documentation updated still makes sense after this update, so let me know if something has become obsolete.


src/perl/Bio/EnsEMBL/Pipeline/PipeConfig/BRC4_genome_loader_conf.pm
```diff
@@ -350,12 +339,11 @@ sub pipeline_analyses {
       -module      => 'Bio::EnsEMBL::Hive::RunnableDB::SystemCmd',
       -parameters  => {
         cmd => 'mkdir -p #log_path#; '
-             . 'echo "checking #json# against #schema#" > #log_path#/#metadata_type#.log; '
-             . 'schemas_json_validate --json_file #json# --json_schema #schema# '
+             . 'echo "checking #json# against #metadata_type#" > #log_path#/#metadata_type#.log; '
+             . 'schemas_json_validate --json_file #json# --json_schema #metadata_type# '
```
vsitnik, line 343: just a note.`metadata_type` -> `JSON file` mapping is done in `schemas_json_validate` itself https://github.com/Ensembl/ensembl-genomio/blob/fb33daabb7502fb25b240654988c506c7ddec2de/src/python/ensembl/io/genomio/schemas/json/validate.py#L45

JAlvarezJarreta: Do you want me to add this comment in the code? Or just leaving this here as a future reference.

vsitnik: m.b. we need this in the documentation

src/python/ensembl/io/genomio/schemas/json/validate.py
```diff
@@ -12,43 +12,70 @@
...
 import jsonschema
+import importlib_resources
 
 from ensembl.utils.argparse import ArgumentParser
 
 
-def json_schema_validator(json_file: PathLike, json_schema: PathLike) -> None:
+_JSON_SCHEMAS = {
+    x.stem: x
+    for x in importlib_resources.files("ensembl.io.genomio.data.schemas").iterdir()
```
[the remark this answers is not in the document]
JAlvarezJarreta: This refers to the data files themselves, not the schema JSON files, so I believe it is still valid.

Regarding your first comment, `import_resources` should only be used to import data files linked to a library. I have already added an option in the function to pass a paths for other schemas. Let me know if I'm missing something else here.

pipelines/nextflow/modules/schema/check_json_schema.nf
```diff
@@ -24,8 +24,8 @@ process CHECK_JSON_SCHEMA {
         tuple val(meta), path(json_file), emit: verified_json
 
     shell:
-        schema = params.json_schemas[json_file.baseName]
+        schema_name = json_file.baseName
```
vsitnik, line 27:
```suggestion
        schema_name = json_file.simpleName
```
mb use `getSimpleName` instead of `getBaseName `  to exclude file extension as well? 
https://www.nextflow.io/docs/latest/script.html#getting-file-attributes

vsitnik (commented):
fine, but need to check if we should have `simpleName` instead of `baseName`

pipelines/nextflow/modules/schema/check_json_schema.nf
JAlvarezJarreta: Since there is only one extension, `baseName` works for this case, but I feel safer with `simpleName`, cheers!

schemas/README.md
```diff
@@ -1,39 +0,0 @@
...
-
-eval "$(pyenv init -)"
-```
-
-Install jsonschema:
-```
-pip install jsonschema
-```
-
-To validate JSON with schema just load it with python:
```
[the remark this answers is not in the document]
JAlvarezJarreta: This is the only file I was not sure what to do with it: in one hand when reading it I felt it was not that useful (we are already providing a method to validate a file with a JSON schema) and there is no information about how to produce such schemas (which I thought that is what it was about), but on the other if it was there is because maybe it was useful for some of you. I'm happy to bring it back in if requested.

MatBarba (approved):

schemas/README.md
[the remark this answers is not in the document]
MatBarba: I believe if someone wants to use json schema to validate files that we do not support here, then they just need to look at the Python json schema documentation itself? No need to do that ourselves here.

vsitnik (approved):
ok with me.
Perhaps, we need an option to allow for validation of any "external" json schemas. But this can be done with a later PR.

schemas/README.md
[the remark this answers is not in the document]
vsitnik: for me this was a useful hack I'm always forgetting how to run.
M.b. we can point to our own validator script usage. Or just have it on our docs  🙈

PR #228, opened 2023-11-27
update schema call for dumpers

MatBarba:
Couple of simple fixes for the dumpers


JAlvarezJarreta (approved):
Cheers! These two slipped through the several checks I did on the codebase :see_no_evil:

PR #229, opened 2023-11-28
replace when by if, clearer structure

MatBarba:
- Use if in the workflows instead of when in the processes
- Pass the selection of files to dump directly instead of as part of the db metadata

### DIFF
--- a/pipelines/nextflow/subworkflows/dump_files/main.nf
+++ b/pipelines/nextflow/subworkflows/dump_files/main.nf
@@ -64,7 +64,6 @@ workflow DUMP_FILES {
         // }
         
         // Events
-
         if ("events" in selection) {
             events = DUMP_EVENTS(db)
             db_files = db_files.concat(events)
--- a/pipelines/nextflow/subworkflows/dump_files/main.nf
+++ b/pipelines/nextflow/subworkflows/dump_files/main.nf
@@ -30,6 +30,8 @@ include { PUBLISH_DIR } from '../../modules/files/publish_output_dump.nf'
 workflow DUMP_FILES {
     take:
         db
+        selection
+        selection_number
 
     emit:
         db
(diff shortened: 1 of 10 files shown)


pipelines/nextflow/subworkflows/dump_files/main.nf
```diff
@@ -38,41 +40,56 @@ workflow DUMP_FILES {
...
-        // db_files = db_files.concat(gff3)
+        // if ("gff3" in selection) {
+            // gff3 = DUMP_GFF3(db)
+            // db_files = db_files.concat(gff3)
+        // }
         
         // Events
-        events = DUMP_EVENTS(db)
-        db_files = db_files.concat(events)
+
```
JAlvarezJarreta, line 67:
```suggestion
```

pipelines/nextflow/workflows/dumper_pipeline/main.nf
```diff
@@ -106,15 +106,13 @@ if (params.host && params.port && params.user && params.output_dir) {
...
 
 // Select the files to dump
-dump_meta = default_selection
-dump_sql = true
+dump_selection = default_selection
 dump_number = 8
 if (params.select_dump) {
-    dump_meta = []
-    dump_sql = false
+    dump_selection = []
```
JAlvarezJarreta, line 112:
```suggestion
```
Since you are assigning it a new value below I don't think this is needed.

JAlvarezJarreta (approved):
Good update :+1:

pipelines/nextflow/workflows/dumper_pipeline/main.nf
```diff
@@ -106,15 +106,12 @@ if (params.host && params.port && params.user && params.output_dir) {
 }
 
 // Select the files to dump
-dump_meta = default_selection
-dump_sql = true
+dump_selection = default_selection
 dump_number = 8
```
Dishalodha, line 110: If we do not include gff, should this be 7 or I am missing something ?

PR #230, opened 2023-11-28
Rewrite dumper parameters

MatBarba:
One last change to make things clearer I hope:
- Separate choices for SQL and other files: --dump_sql --dump_all_files or --dump_selection
- List of files and their numbers is in one simple map

### DIFF
--- a/pipelines/nextflow/workflows/dumper_pipeline/main.nf
+++ b/pipelines/nextflow/workflows/dumper_pipeline/main.nf
@@ -64,7 +64,7 @@ def helpMessage() {
         nextflow run \\
             -w \${data_dir}/nextflow_work \\
             ensembl-genomio/pipelines/nextflow/workflows/dumper_pipeline/main.nf \\
-            --dump_sql --dump_files \\
+            --dump_sql --dump_all_files \\
             -profile lsf \\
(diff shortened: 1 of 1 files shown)


JAlvarezJarreta (approved):
Looks good to me. :+1: 

The only comment (that I have noticed now) is that it may be wiser to declare `dump_selection` as `param.dump_selection` and use it directly in the subworkflow instead of passing it, as I have found this "passing a variable instead of a channel" sometimes dangerous and leading to hard-to-notice bugs.

MatBarba:
> Looks good to me. 👍
> 
> The only comment (that I have noticed now) is that it may be wiser to declare `dump_selection` as `param.dump_selection` and use it directly in the subworkflow instead of passing it, as I have found this "passing a variable instead of a channel" sometimes dangerous and leading to hard-to-notice bugs.

I'm kind of thinking the opposite? I'm trying to avoid using too many params values as global parameters, i.e. we should be able to use the subworkflows by providing all info necessary directly instead of -- sometimes hard to find -- params values.

(One example where that would cause a problem is when we call a subworkflow several times in the same workflow).

Dishalodha (approved):

PR #232, opened 2023-11-29
Mbarba/less check procs

MatBarba:
Simply check the json files when they are created: this means less jobs to run, and if something is wrong the generator process is the one that fails.

NB: not possible to do the same with the gff3 validation, because it uses `gt` from a container, and I don't think it's possible to have both the current environment (e.g. to use `gff3_process`) and a container in the same process script (or at least, not as simply).

### DIFF
--- a/src/python/ensembl/io/genomio/seq_region/prepare.py
+++ b/src/python/ensembl/io/genomio/seq_region/prepare.py
@@ -435,9 +435,9 @@ def report_to_csv(report_path: PathLike) -> Tuple[str, Dict]:
 
 
 def prepare_seq_region_metadata(
+    dst_file: PathLike,
     genome_file: PathLike,
     report_file: PathLike,
-    dst_dir: PathLike,
     gbff_file: Optional[PathLike] = None,
     brc_mode: bool = False,
     to_exclude: Optional[List[str]] = None,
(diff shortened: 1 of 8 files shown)


src/python/ensembl/io/genomio/seq_region/prepare.py
```diff
@@ -435,9 +435,9 @@ def report_to_csv(report_path: PathLike) -> Tuple[str, Dict]:
 
 
 def prepare_seq_region_metadata(
+    dst_file: PathLike,
```
JAlvarezJarreta, line 438: I would still prefer to have this after the input files, to follow the usual `src`, `dst` order, but this is just a preference.

JAlvarezJarreta (approved):
As mentioned before, I quite like including the check inside the process :+1:

src/python/ensembl/io/genomio/seq_region/prepare.py
JAlvarezJarreta: @MatBarba forgot to move this one as well with your [last commit](https://github.com/Ensembl/ensembl-genomio/pull/232/commits/20c95f363ee5cb2ec0f61e1d6548ccfb3d15b7a4).

PR #233, opened 2023-11-30
Do not create an accession subfolder

MatBarba:
Fix a bug from a discrepancy between what the script does (create a subdir within the output dir) and what the Nextflow process expects (no subfolder)


vsitnik (approved):
cool. thank you!

Dishalodha (approved):

MatBarba:
> Thanks Matthieu. Do we have any idea why this just broke recently ? It was working fine for quite some time

My pipeline test did not pick that up because it does not test the dumper and the process at all because of the cache. I think there's a different way to test that (using stub)

PR #240, opened 2023-12-06
Add a db_list file as an option to dump genomes

MatBarba:
Add an option to dump data based on an explicit list of database names from a file.

I've implemented this by transferring the list of database names itself instead of the file through the pipeline: this is to avoid having to stage a file, and it adds flexibility in how we get that list of db names (e.g. we could add an option to dump only one or several dbs from the command line instead of a file).


src/python/ensembl/io/genomio/database/factory.py
```diff
@@ -136,6 +146,10 @@ def main() -> None:
     args = parser.parse_args()
     init_logging_with_args(args)
 
+    db_list_file = None
+    if args.db_list:
+        db_list_file = _load_multine_file(args.db_list)
```
JAlvarezJarreta, line 151: I may be wrong, but if I recall correctly `args.db_list` is always defined, so you could also do:
```suggestion
    if args.db_list:
        args.db_list = _load_multine_file(args.db_list)
```
And use `args.db_list` below.

src/python/ensembl/io/genomio/database/factory.py
```diff
@@ -145,12 +159,15 @@ def main() -> None:
...
     server = CoreServer(server_url)
+    logging.debug("Get databases...")
     databases = server.get_cores(
         prefix=args.prefix,
         build=args.build,
         version=args.version,
         dbname_re=args.db_regex,
+        db_list=db_list_file,
     )
+    logging.info(f"Got {len(databases)} databases")
```
JAlvarezJarreta, line 170:
```suggestion
    logging.info(f"Got {len(databases)} databases")
    logging.debug("\n".join(databases))
```

src/python/ensembl/io/genomio/database/factory.py
MatBarba: Actually `args.db_list` can be empty (in which case the factory will use the filter parameters only)

PR #241, opened 2023-12-06
Seq_region lrg support

MatBarba:
Not sure how to handle this properly: the human core db contains a coord_system_level named `lrg` (not SO terms).

This PR is to make the validation pass for that genome. An alternative would be to add a list of coord_system_level that we would not dump in the seq_region dumper, but that'd mean we don't dump everything.


MatBarba:
> LRGs are quite tricky (the Perl codebase is full with conditionals for this data) and definitely not free of headaches. I agree we should be avoiding these for now. Furthermore, VEuPathDB always claims HostDB species should be basic pointers to better resources such as Ensembl, so this change also aligns with that decision. 👍

This PR does dump the lrg sequences, so do I need to make a different PR where the code explicitly exclude them from the dumps?

JAlvarezJarreta:
Sorry, I completely misread the intention of this PR :man_facepalming: It's ok either way, so if we are currently dumping them without issues, happy with that. :+1:

MatBarba:
Reopening because I don't think it would hurt to have this and it's very simple

PR #242, opened 2023-12-06
Avoid loading the schema from db

MatBarba:
DBConnection loads the whole schema in its __init__, which takes a lot of time and is not actually needed apart from getting meta data. In this PR I override the __init__ without the schema loading (that does make the pylint fail obviously).

NB: this is mostly to accelerate the db_factory (which needs to get metadata for all the databases it lists). Without this fix, the db_factory takes minutes to run (instead of seconds) when it tries to read lots of dbs.

That means the meta checks (schema_type) can't be done that readily, so I've overriden that part for the purpose of my script (that said it should probably not even use the SQLalchemy Meta model here: a raw SQL query might actually be better in the initial code).

This PR is not meant to be merged as is as it is probably not the best solution (though it does work).

### DIFF
--- a/src/python/ensembl/io/genomio/database/core_database.py
+++ b/src/python/ensembl/io/genomio/database/core_database.py
@@ -18,7 +18,7 @@ __all__ = ["CoreDatabase"]
 
 from typing import Dict, List
 
-from sqlalchemy import select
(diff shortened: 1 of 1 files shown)


MatBarba:
Better version: #295

PR #244, opened 2023-12-11
Use new json validator in ehive dumper

MatBarba:
Just replace the `json_validator` with the new version (simpler) in the ehive version for backward compatibility


Dishalodha (approved):
Nice 👌

PR #245, opened 2023-12-12
Update tests tree structure and docstrings

JAlvarezJarreta:
This PR sets all tests (and their content) into the same structure and format, not only internally but as the overall tree folder structure, matching the same structure in the Python library.

_Note:_ I have removed things like `:class:` and such because mkdocs is not capable of processing them, so the result is something like `:class:sqlalchemy.engine` which is neither pretty nor readable.


ens-LCampbell (approved):

MatBarba (approved):
Looks ok. One note: we don't need to `flatfile` subfolder anymore do we?

JAlvarezJarreta:
> Looks ok. One note: we don't need to `flatfile` subfolder anymore do we?

Leaving it there for the time being in case we introduce database folders to create unit test dbs. Can be removed at a later stage easily thanks to your fixtures :relaxed:

PR #247, opened 2023-12-12
Mbarba/hack/gff3 meta

MatBarba:
This PR is to separate the GFFMerger and GFFSimplifier in their own file, as well as getting the list of supported biotypes from a resource file (instead of hardcoded in a module), accessible via a new object, GFFMeta (in it's own file as well).

This is to isolate the various parts and make it easier to test each part. I've added a simple test for the GFFMeta.


JAlvarezJarreta:
One comment before finishing my full review: Why have you introduced `src/python/ensembl/io/genomio/data/__init__.py` (and in the subfolders)? It should not be required, or at least I was unable to find its need anywhere in the documentation and it was working for me without it for `data.schemas`.

MatBarba:
> One comment before finishing my full review: Why have you introduced `src/python/ensembl/io/genomio/data/__init__.py` (and in the subfolders)? It should not be required, or at least I was unable to find its need anywhere in the documentation and it was working for me without it for `data.schemas`.

Good point, I used the wrong guideline for this: importlib_resources does not requires __init__.py files.

MatBarba:
See the unified version https://github.com/Ensembl/ensembl-genomio/pull/254

PR #248, opened 2023-12-12
Mbarba/hack/gff3 annotation tests

MatBarba:
Add tests for the extract_annotation module code


src/python/tests/gff3/test_extract_annotation.py
```diff
@@ -22,15 +22,230 @@
...
+        self, in_parent_type, in_parent_id, in_child_id, out_parent_type, out_child_id, expected
+    ):
+        """Tests the `FunctionaAnnotation.get_parent()` method.
+
+        Args:
+            in_parent_type: Type for the SeqFeature parent.
+            in_parent_id: ID for the SeqFeature parent.
+            in_child_id: ID for the SeqFeature child.
+            out_parent_type: Type for the parent stored in the FunctionalAnnotation.
+            out_parent_id: ID for the parent stored.
```
JAlvarezJarreta, line 116:
```suggestion
            out_child_id: ID for the child stored.
```

src/python/tests/gff3/test_extract_annotation.py
```diff
@@ -22,15 +22,230 @@
...
+            out_parent = annot.get_parent(out_parent_type, out_child_id)
+            assert out_parent == in_parent_id
+
+    @pytest.mark.parametrize(
+        "child_type, child_id, out_parent_id, expected",
+        [
+            ("transcript", "mrna_A", "gene_A", does_not_raise()),
+            ("bad_type", "mrna_A", "gene_A", raises(KeyError)),  # Child type does not exist
+            ("gene", "gene_A", None, raises(AnnotationError)),  # Feature ID already loaded
+            ("gene", "gene_B", "gene_A", raises(AnnotationError)),  # Cannot add a gene child of a gene
```
JAlvarezJarreta, line 137: Alternative proposal:
```suggestion
            ("bad_type", "mrna_A", "gene_A", raises(KeyError), id="unknown_child_type"),
            ("gene", "gene_A", None, raises(AnnotationError), id="feature_id_already_loaded"),
            ("gene", "gene_B", "gene_A", raises(AnnotationError), id="circular_gene_addition"),
```

src/python/tests/gff3/test_extract_annotation.py
```diff
@@ -22,15 +22,230 @@
...
+            child_id: ID for the SeqFeature child.
+            out_parent_id: ID for the parent.
+            expected: What exception is expected to be raised, if any.
+
+        """
+        annot = FunctionalAnnotations()
+        with expected:
+            parent = SeqFeature(type="gene", id="gene_A")
+
+            child = SeqFeature(type="mRNA", id=child_id)
```
JAlvarezJarreta, line 158: Same as before.
```suggestion
            parent = SeqFeature(type="gene", id="gene_A")
            child = SeqFeature(type="mRNA", id=child_id)
```

src/python/tests/gff3/test_extract_annotation.py
```diff
@@ -22,15 +22,230 @@
...
+        """Tests the `FunctionaAnnotation.add_feature()` method with only one feature.
+
+        Args:
+            seq_feat_type: Type for the SeqFeature to add.
+            feat_type: Type category for that SeqFeature.
+
+        """
+        annot = FunctionalAnnotations()
+        with expected:
+            feature = SeqFeature(type=seq_feat_type, id="featA")
```
JAlvarezJarreta, line 67: Avoid having code that does not belong to your codebase inside the context manager to avoid an unexpected exception making your test pass, e.g. if this line was to raise a `KeyError` in your last parametrization the test would pass successfully without even running the `add_feature()` part.

src/python/tests/gff3/test_extract_annotation.py
```diff
@@ -22,15 +22,230 @@
...
+        Args:
+            feat_type: Type for the features to fetch.
+            expected: What exception is expected to be raised, if any.
+
+        """
+        annot = FunctionalAnnotations()
+        one_gene = SeqFeature(type="gene", id="gene_A")
+        one_transcript = SeqFeature(type="mRNA", id="mrna_A")
+        annot.add_feature(one_gene, "gene")
+        annot.add_feature(one_transcript, "transcript", parent_id=one_gene.id)
```
JAlvarezJarreta, line 185: Do we want to create a dependency so this test is not run if `add_feature()` tests don't pass?

src/python/tests/gff3/test_extract_annotation.py
```diff
@@ -22,15 +22,230 @@
...
+
+        with expected:
+            out_feats = annot.get_features(feat_type)
+            expected_number = 0
+            if feat_type in ("gene", "transcript"):
+                expected_number = 1
```
JAlvarezJarreta, line 191: I would add `expected_number` as a parameter of the test to add more flexibility.

JAlvarezJarreta (changes requested):
The PR looks really good and it involves the testing of something rather complex with several object types. Well done. A few suggestions:
* All the new `raise KeyError()` elements in `src/python/ensembl/io/genomio/gff3/extract_annotation.py` could be changed from an `if-else` block to `try-except` to align better with Python style. Furthermore, if you believe the default message in the exception is good enough (I believe it would just say the key "parent-id" is missing), we could remove these altogether.
* I believe it may be best to opt of the declaration of the `SeqFeatures` needed in the parametrization rather than inside the code, as this would simplify the test code (less ifs and such). The drawback is that you will need to add `id=` to all parametrizations to make them readable in the terminal (not the end of the world in my opinion)
    * This combined with a list argument may allow you to merge the two `test_add_feature_*` methods (although I'm not 100% sure)

src/python/ensembl/io/genomio/gff3/extract_annotation.py
```diff
@@ -154,7 +156,7 @@ def _generic_feature(self, feature: SeqFeature, feat_type: str) -> Dict[str, Any
 
         return feature_object
 
-    def _transfer_descriptions(self) -> None:
+    def transfer_descriptions(self) -> None:
```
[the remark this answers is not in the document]
MatBarba: I used that method in the `test_*` file, but mypy wasn't too happy that I did so with a private method...

MatBarba:
See the unified version in https://github.com/Ensembl/ensembl-genomio/pull/254

PR #252, opened 2024-01-03
Add exclusion directory list to license header checks

JAlvarezJarreta:
Test files should be minimal and normally these do not require the license header to be attached to them, so we can relax this requirement for such files.


MatBarba (approved):
Works as intended. We should move the Python test data files in the `src/python/tests/data subfolder` at some point

PR #256, opened 2024-01-09
Remove GFFMeta class

JAlvarezJarreta:
Alternative (and simpler) proposal to the GFFMeta class. Similar properties than #255 but the amount of lines involved is reduced whilst the readability is preserved.

_Note:_ I have run `pytest` and every test succeeds.


MatBarba (approved):
Does roughly the same thing without the added complexity

PR #260, opened 2024-01-11
updating the container as we require ps for tracing

Dishalodha:
Updating the container so trace parameter can run smoothly


JAlvarezJarreta (approved):

ens-LCampbell:
Not much to review in commit. Is there any definition file or is this a public image with ps built in ?

MatBarba (approved):

ens-LCampbell (approved):
All good from me

PR #269, opened 2024-01-17
Mbarba/hack/fix fasta tests

MatBarba:
Fix fasta tests to use the new `data_dir` fixture:
* Move the data files in a subdir with the same name as the test file
* Use the `data_dir` that points to that subfolder

Other tweaks to simplify the code:
* Remove the setup
* Remove the test class


src/python/tests/fasta/test_fasta_prep_data.py
```diff
@@ -22,120 +22,113 @@
...
 from pathlib import Path
-from typing import ContextManager, List, Set
+from typing import ContextManager, Set
 
 import pytest
 
 import ensembl.io.genomio.fasta.process as FastaProcessing
 
 
-class TestFastaProcess:
```
ens-LCampbell, line 32: I thought I was following best practice to implement via a class. Any one explain why its better to remove said class ?

ens-LCampbell (approved):
Assuming this was tested ? Happy with these changes thank you @MatBarba

src/python/tests/fasta/test_fasta_prep_data.py
ens-LCampbell: Jorge Quote -> "So whilst the class approach is the best one when a setup is required for a set of tests, in your case you can use the two fixtures directly in your tests without the need of additional configuration. Thus why it can be removed. But I would suggest to always start with a class, and only remove it when you realise it is not required."

PR #270, opened 2024-01-17
Mbarba/fix integrity

MatBarba:
A silent bug in the integrity check where the GFF3 comparison was not done because we used wrong keys.

I've fixed it and added some failchecks.

### DIFF
--- a/src/python/ensembl/io/genomio/manifest/check_integrity.py
+++ b/src/python/ensembl/io/genomio/manifest/check_integrity.py
@@ -78,22 +78,32 @@ class Manifest:
         self.brc_mode = False
 
     def has_lengths(self, name: str) -> bool:
-        """Check if a given name has lengths records."""
-        if name in self.lengths:
-            return True
-        return False
+        """Check if a given name has lengths records.
+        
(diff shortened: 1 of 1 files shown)


src/python/ensembl/io/genomio/manifest/check_integrity.py
```diff
@@ -78,22 +78,32 @@ def __init__(self, manifest_path: PathLike) -> None:
...
-        return False
+        """Check if a given name has lengths records.
+        
+        Raise KeyError if the name is not supported.
+
+        """
+        try:
+            if self.lengths[name]:
+                return True
+            return False
```
JAlvarezJarreta, line 89: Just to double-check: if `self.lengths[name]` exists but the dictionary inside is empty, `False` will be returned as well. This was the behaviour before, but want to confirm that is what we want.

JAlvarezJarreta (approved):
Good catch :+1:

src/python/ensembl/io/genomio/manifest/check_integrity.py
MatBarba: That's correct

PR #273, opened 2024-01-18
Fix GFF3 parsing for naked transcripts and CDSs

MatBarba:
The GFF3 part of the hackthon branch introduce some changes that break how features are handled (because of a change in the order of operations). This is some code in the simplifier module which doesn't have tests yet.

Details:
In the current (faulty) order of operation, the features that have a biotype mRNA, transcripts, etc. or CDS without a gene parent are deemed unsupported and the GFF3 parsing fails. We do have two operations `transcript_gene()` and `CDS_gene()` that move the mRNA, CDS etc. under a gene parent, but right now they are done after checking for the biotype.

In this patch I reorder the operations with clearer labels, and put back the two operations that add a gene parent before the check for unsupported gene features.

NB: I recommend looking at the full simpler_gff3_feature method before and after, instead of the line by line diff.

### DIFF
--- a/src/python/ensembl/io/genomio/data/gff3/biotypes.json
+++ b/src/python/ensembl/io/genomio/data/gff3/biotypes.json
@@ -35,6 +35,7 @@
     },
     "non_gene": {
         "supported": [
+            "mobile_genetic_element",
             "transposable_element"
         ]
     },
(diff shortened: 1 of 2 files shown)


JAlvarezJarreta (approved):
I lack the knowledge to understand the reason behind the reordering completely, but the code changes definitely make sense and look good :+1:

PR #275, opened 2024-01-23
Make the Python image a CI/CD variable

JAlvarezJarreta:
Instead of setting only the version as the environment variable content I have opted for the whole image information in case we want to use a different source in the future.


vsitnik (approved):
cool. should work
https://docs.gitlab.com/ee/ci/variables/#cicd-variable-precedence

PR #276, opened 2024-01-23
Update stable id pattern in OSID

MatBarba:
This script can now be used to update the pattern of stable_ids stored in OSID: it can do so as long as we do not also include start for gene and transcript (because in this case it will not work, otherwise it would change what has already been allocated).

With some clean up.

### DIFF
--- a/scripts/brc4/patch_build/osid_add_organisms.py
+++ b/scripts/brc4/patch_build/osid_add_organisms.py
@@ -14,43 +14,54 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Add or update one organism in an OSID server.
+
+You can update an organism by setting a new pattern without setting the gene and transcript starts.
+"""
+
 
 import argparse
-import requests
 import json
+from typing import Dict, Optional
+
+import requests
 
 
-def get_organism(url, user, key, org_data):
+def get_organism(url: str, user: str, key: str, org_data: Dict[str, str]) -> Optional[int]:
+    """Retrieve the organism ID from OSID for a given organism.
+
+    Returns None if no organism could be found with that name.
+    """
     species = org_data["organismName"]
 
     body = {"organismName": species}
 
     get_org = url + "/organisms"
-    result = requests.get(get_org, auth=(user, key), params=body)
+    result = requests.get(get_org, auth=(user, key), params=body, timeout=10)
 
     if result and result.status_code == 200:
         organisms = json.loads(result.content)
-        count = len(organisms)
 
         if len(organisms) == 1:
             return organisms[0]["organismId"]
-        elif len(organisms) == 0:
-            return False
-        else:
-            raise Exception("Found several organisms for the name '%s'" % species)
-    else:
-        print("Error: " + str(result))
+        if len(organisms) == 0:
+            return None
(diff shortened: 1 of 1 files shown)


scripts/brc4/patch_build/osid_add_organisms.py
```diff
@@ -14,38 +14,54 @@
...
-        elif len(organisms) == 0:
-            return False
-        else:
-            raise Exception("Found several organisms for the name '%s'" % species)
-    else:
-        print("Error: " + str(result))
+        if len(organisms) == 0:
+            return None
+        raise ValueError(f"Found several organisms for the name '{species}'")
+    print("Error: " + str(result))
```
JAlvarezJarreta, line 50:
```suggestion
    print(f"Error: {result}")
```

scripts/brc4/patch_build/osid_add_organisms.py
```diff
@@ -14,38 +14,54 @@
...
         if len(organisms) == 1:
             return organisms[0]["organismId"]
-        elif len(organisms) == 0:
-            return False
-        else:
-            raise Exception("Found several organisms for the name '%s'" % species)
-    else:
-        print("Error: " + str(result))
+        if len(organisms) == 0:
+            return None
```
JAlvarezJarreta, line 48: Really looking forward to Python 3.10: https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching

scripts/brc4/patch_build/osid_add_organisms.py
```diff
@@ -14,38 +14,54 @@
...
+    print("Error: " + str(result))
+    return None
+
 
+def add_organism(url: str, user: str, key: str, org_data: Dict[str, str]) -> None:
+    """Add a new organism to OSID."""
+
+    # Species is new, check we have all the keys
+    expected = {"organismName", "template", "geneIntStart", "transcriptIntStart"}
+    if set(expected) != set(org_data):
```
JAlvarezJarreta, line 59:
```suggestion
    if set(expected) != org_data.keys():
```
Equivalent code (tested) and, in my opinion, more readable since it is not clear what `set()` is using from the dictionary.

scripts/brc4/patch_build/osid_add_organisms.py
```diff
@@ -54,11 +70,13 @@ def add_organism(url, user, key, org_data):
         print(json.dumps(json.loads(result.content), indent=4))
 
 
-def update_organism(url, user, key, org_data, organism_id):
+def update_organism(url: str, user: str, key: str, org_data: Dict[str, str], organism_id: int):
```
JAlvarezJarreta, line 73:
```suggestion
def update_organism(url: str, user: str, key: str, org_data: Dict[str, str], organism_id: int) -> None:
```

JAlvarezJarreta (approved):
Looks good to me. Nice to see this new feature. Couple of details to address, but nothing major/critical.

Dishalodha (approved):

scripts/brc4/patch_build/osid_add_organisms.py
MatBarba: I agree, but I would still prefer to keep the set() on the right side, because it's not clear what comparing a set and a list should output

PR #279, opened 2024-01-25
Split up ID allocation code and test

MatBarba:
I've moved all the methods about IDs checking and allocation into a separate module/object.

With various tests, of basically 3 kinds:
- Simple unit tests (string -> string)
- Tests that load data from a GFF3 and check a given output (e.g. generate an ID from a GFF3 feature)
- Test that load data from a GFF3, then write it back out to compare with an expected GFF3 file

Note that using GFF3 files instead of instanciating SeqRecords/SeqFeatures is needed to avoid the dependencies issues between Biopython and the BCbio GFF3 parser. It should also be more robust when other changes/deprecations might occur down the line.


JAlvarezJarreta (changes requested):
A much needed code refactoring :+1:

Some of the suggestions are focused on better readability/code homogeneity. I have suggested to make a couple of methods static based on their current implementation. If you like/approve the suggestion of renaming the class to `StableIDAllocator`, I guess we should also change the python file name accordingly.

I believe `validate_gene_id` should be renamed to `skip_gene_id_validation` and change the logic in `is_valid()`, as it makes the code more readable (in my opinion). However, I haven't suggested this change as I wasn't 100% I have understood its purpose fully.

src/python/tests/gff3/test_id_allocator.py
```diff
@@ -0,0 +1,225 @@
...
+) -> None:
+    """Test pseudogene CDS ID normalization."""
+    ids = IDAllocator()
+
+    # Load record and update feature
+    record = _read_record(data_dir / input_gff)
+    features = record.features
+    record.features = []
+    for feature in features:
+        ids.normalize_pseudogene_cds_id(feature)
```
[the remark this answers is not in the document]
MatBarba: That method name should not change (we are indeed updating the IDs of the CDSs of the pseudogene)

src/python/tests/gff3/test_id_allocator.py
```diff
@@ -0,0 +1,225 @@
...
+@pytest.mark.parametrize(
+    "prefix, expected_ids",
+    [
+        pytest.param(None, ["TMP_1", "TMP_2"], id="Default prefix"),
+        pytest.param("MYPREF_", ["MYPREF_1", "MYPREF_2"], id="Prefix MYPREF_"),
+    ],
+)
+def test_generate_id(prefix: str, expected_ids: List[str]) -> None:
+    """Test IDs generation."""
+    ids = IDAllocator()
```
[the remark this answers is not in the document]
MatBarba: Test updated, we are testing with no default prefix change, and with no prefix aka an empty string

MatBarba:
> A much needed code refactoring 👍
> 
> Some of the suggestions are focused on better readability/code homogeneity. I have suggested to make a couple of methods static based on their current implementation. If you like/approve the suggestion of renaming the class to `StableIDAllocator`, I guess we should also change the python file name accordingly.
> 
> I believe `validate_gene_id` should be renamed to `skip_gene_id_validation` and change the logic in `is_valid()`, as it makes the code more readable (in my opinion). However, I haven't suggested this change as I wasn't 100% I have understood its purpose fully.

I've made the validation flag change, I think that makes sense

src/python/ensembl/io/genomio/gff3/id_allocator.py
```diff
@@ -0,0 +1,191 @@
...
+from dataclasses import dataclass, field
+import logging
+import re
+from typing import List, Set
+
+from Bio.SeqFeature import SeqFeature
+
+
+class InvalidStableID(ValueError):
+    """Raised when there is a problem with an ID."""
```
JAlvarezJarreta, line 28:
```suggestion
    """Raised when there is a problem with an stable ID."""
```

src/python/ensembl/io/genomio/gff3/id_allocator.py
```diff
@@ -0,0 +1,191 @@
...
+        if len(stable_id) < self.min_id_length:
+            logging.debug(f"Stable id is too short (<{self.min_id_length}) {stable_id}")
+            return False
+
+        return True
+
+    @staticmethod
+    def remove_prefix(stable_id: str, prefixes: List[str]) -> str:
+        """Returns the stable ID after removing its prefix (if any).
+        Args:
```
JAlvarezJarreta, line 88:
```suggestion

        If more than one prefix may be found, only the first one is removed.

        Args:
```

src/python/ensembl/io/genomio/gff3/id_allocator.py
```diff
@@ -0,0 +1,191 @@
...
+
+        """
+        if number < 1:
+            raise ValueError("Number has to be a positive integer.")
+
+        transcript_id = f"{gene_id}_t{number}"
+        return transcript_id
+
+    def normalize_cds_id(self, cds_id: str) -> str:
+        """Returns a normaliSed version of the provided CDS ID.
```
JAlvarezJarreta, line 115:
```suggestion
        """Returns a normalized version of the provided CDS ID.
```
Since you are using the American spelling elsewhere.

src/python/ensembl/io/genomio/gff3/simplifier.py
```diff
@@ -87,7 +85,17 @@ def __init__(self, genome_path: Optional[PathLike] = None, make_missing_stable_i
         if genome_path:
             with Path(genome_path).open("r") as genome_fh:
                 self.genome = json.load(genome_fh)
-        self.make_missing_stable_ids: bool = make_missing_stable_ids
+        self._set_id_prefix()
+
+    def _set_id_prefix(self) -> None:
+        """Set the ID prefix using the organism abbrev if it exists in the genome metadata."""
```
JAlvarezJarreta, line 91:
```suggestion
        """Sets the ID prefix using the organism abbrev if it exists in the genome metadata."""
```

src/python/tests/gff3/test_id_allocator.py
```diff
@@ -0,0 +1,250 @@
...
+def _read_record(in_gff: Path) -> SeqRecord:
+    """Get one clean record from a GFF3 file."""
+    with in_gff.open("r") as in_gff_fh:
+        for record in GFF.parse(in_gff_fh):
+            new_record = SeqRecord(seq=record.seq, id=record.id)
+            new_record.features = record.features
+            return new_record
+
+    raise ValueError(f"No feature loaded from {in_gff}")
+
```
JAlvarezJarreta, line 40: Do we want to make this a fixture as we have done for the automatic load of JSON files?

src/python/ensembl/io/genomio/gff3/id_allocator.py
```diff
@@ -0,0 +1,191 @@
...
+class StableIDAllocator:
+    """Set of tools to check and allocate stable IDs."""
+
+    # Multiple parameters to automate various fixes
+    skip_gene_id_validation: bool = False
+    min_id_length: int = 7
+    current_id_number: int = 0
+    make_missing_stable_ids: bool = True
+    prefix: str = "TMP_"
+    _loaded_ids: Set = field(default_factory=set)
```
JAlvarezJarreta, line 41: Quite a neat trick! :nerd_face:

src/python/ensembl/io/genomio/gff3/id_allocator.py
```diff
@@ -0,0 +1,191 @@
...
+                    new_gene_id = new_gene_id_base
+                    number = 1
+                    while new_gene_id in self._loaded_ids:
+                        number += 1
+                        new_gene_id = f"{new_gene_id_base}_{number}"
+                        if number > 10:
+                            raise InvalidStableID(f"Duplicate ID {new_gene_id_base} (up to {new_gene_id})")
+                    self._loaded_ids.add(new_gene_id)
+                    logging.debug(f"Using GeneID {new_gene_id} for stable_id instead of {gene.id}")
+                    return new_gene_id
```
JAlvarezJarreta, line 182: I'm struggling to understand the usage of `_loaded_ids`, in particular why you only add them here and not when a new `gene_id` is generated in the method above.

JAlvarezJarreta (approved):
Great second round, almost there.

I have left a comment about `StableIDAllocator._loaded_ids` because I'm struggling to understand its functionality (more in the comment). Besides that, good to go!

src/python/ensembl/io/genomio/gff3/id_allocator.py
MatBarba: Yeah, this is to avoid setting the default as just `set()` which has unintended consequences if we use more than 1 instance of the class

src/python/tests/gff3/test_id_allocator.py
MatBarba: Yeah I plan to do that in a new PR

PR #284, opened 2024-02-06
Add tests for GFF3 merger

MatBarba:
Add more test cases for GFF3 merge with better tests and files names.

Adds a custom diff to show the difference between 2 files when they differ


JAlvarezJarreta (approved):
Looks good. A couple of comments, but the suggestion can be done in a separate PR.

PR #288, opened 2024-02-07
Don't fail but write a stats error file

MatBarba:
Instead of dying when stats do not match, simply create a stats error file that we can easily see after the script (and the prepare genome pipeline) is over.

### DIFF
--- a/src/python/ensembl/io/genomio/manifest/compute_stats.py
+++ b/src/python/ensembl/io/genomio/manifest/compute_stats.py
@@ -70,7 +70,7 @@ class manifest_stats:
         self.manifest = f"{manifest_dir}/manifest.json"
         self.accession: Optional[str] = accession
         self.errors: List[str] = []
-        self.errors_file = f"{manifest_dir}/stats_diff.json"
+        self.errors_file = Path(manifest_dir) / "stats_diff.log"
         if datasets_bin is None:
             datasets_bin = "datasets"
         self.datasets_bin = datasets_bin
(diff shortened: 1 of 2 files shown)


src/python/ensembl/io/genomio/manifest/compute_stats.py
```diff
@@ -69,7 +69,8 @@ class manifest_stats:
     def __init__(self, manifest_dir: str, accession: Optional[str], datasets_bin: Optional[str]):
         self.manifest = f"{manifest_dir}/manifest.json"
         self.accession: Optional[str] = accession
-        self.error = False
+        self.errors: List[str] = []
+        self.errors_file = f"{manifest_dir}/stats_diff.json"
```
JAlvarezJarreta, line 73:
```suggestion
        self.errors_file = Path(manifest_dir) / "stats_diff.log"
```
I may be wrong, but it looks like you are printing simple text inside this file.

src/python/ensembl/io/genomio/manifest/compute_stats.py
```diff
@@ -101,8 +102,10 @@ def run(self, stats_path: PathLike):
             stats_out.write("\n".join(stats))
 
         # Die if there were errors in stats comparison
-        if self.error:
-            raise StatsError(f"Stats count errors, check the file {stats_path}")
+        if self.errors:
+            with open(self.errors_file, "w") as errors_fh:
```
JAlvarezJarreta, line 106:
```suggestion
            with self.errors_file.open("w") as errors_fh:
```
Just to use the `Path()` features.

JAlvarezJarreta (approved):
Looks good to me :+1: The only relevant bit of my review is the `errors_file` extension, the other part is just a preference.

src/python/ensembl/io/genomio/manifest/compute_stats.py
MatBarba: Indeed, good catch

PR #289, opened 2024-02-08
Create assert_files fixture

MatBarba:
Compare two files and show a unified diff if they do differ. Pytest output looks like this if the assertion fails:

```
AssertionError: Test-made and expected files differ
  --- Test-made file
   +++ Expected file
   @@ -1,3 +1,3 @@
    A
    -B1
    +B2
    C  
assert False
```

[...]

```
FAILED test_xxx.py::test_xxx[name] - AssertionError: Test-made and expected files differ
```

### DIFF
--- a/src/python/tests/conftest.py
+++ b/src/python/tests/conftest.py
@@ -20,7 +20,6 @@ will have access to the plugins, hooks and fixtures defined here.
 """
 
 from difflib import unified_diff
-import filecmp
 from pathlib import Path
 from typing import Any, Callable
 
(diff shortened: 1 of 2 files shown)


src/python/tests/conftest.py
```diff
@@ -64,3 +66,19 @@ def _json_data(file_name: str) -> Any:
         return get_json(data_dir / file_name)
 
     return _json_data
+
+
+@pytest.fixture(name="assert_files")
+def assert_files() -> Callable[[Path, Path], None]:
```
JAlvarezJarreta, line 72:
```suggestion
def assert_files() -> Callable:
```
Just a suggestion, but I believe it may make things less complicated (and long).

JAlvarezJarreta (approved):
I would have never considererd wrapping the whole assert, nice one!

PR #291, opened 2024-02-13
download_single_file mock suggestion

MatBarba:
One way I made it work: I mocked `FTP` and replaced its `retrbinary()` method by a function that copies the expected file content


ens-LCampbell (approved):

ens-LCampbell:
I have added these suggestions manually, and to avoid messing things up I will just close this PR. But all suggestions will be retained thanks @MatBarba

PR #292, opened 2024-02-13
Rename BRC seq_region names

MatBarba:
A script to automate the renaming of seq_region `BRC4_seq_region_name` attributes.

The script accepts a simple tab file with the current name and the new BRC name, and checks against the current database to see what it would do (not found, insert, update, or already exists). Then just add the usual `--update` argument to make the changes.

NB: the current name can be the seq_region name or a synonym.


src/python/ensembl/io/genomio/seq_region/rename.py
```diff
@@ -0,0 +1,212 @@
...
+def get_rename_map(map_file: Path) -> List[SeqRegionReplacement]:
+    """Load requested renaming from a tab file."""
+
+    seq_regions: List[SeqRegionReplacement] = []
+    with map_file.open("r") as input_fh:
+        for line in input_fh:
+            try:
+                (name, brc_name) = [item.strip() for item in line.split("\t")]
+            except ValueError:
+                logging.warning(f"Skipping line '{line.strip()}', wrong format")
```
JAlvarezJarreta, line 67: Quite smart, I like it! :nerd_face:

src/python/ensembl/io/genomio/seq_region/rename.py
```diff
@@ -0,0 +1,212 @@
...
+def get_seq_regions_to_replace(
+    session: Session, rename_map: List[SeqRegionReplacement]
+) -> List[SeqRegionReplacement]:
+    """Check each seq_region replacement against the database.
+
+    Args:
+        session: Session from the current core.
+        rename_map: List of remappings to check.
+
+    Returns: same list with the operation changed to the operation it can do.
```
JAlvarezJarreta, line 84:
```suggestion
    Returns:
        Same list of remappings with the operation changed to the operation it can do.
```

src/python/ensembl/io/genomio/seq_region/rename.py
```diff
@@ -0,0 +1,212 @@
...
+from ensembl.utils.logging import init_logging_with_args
+
+
+class Operation(Enum):
+    """Controlled list of operations for the seq_region replacement."""
+
+    do_nothing = auto()
+    update = auto()
+    insert = auto()
+    not_found = auto()
```
JAlvarezJarreta, line 38: Due to their nature of being constants, `Enum` attributes are strongly recommended to be written in upper case, e.g. `DO_NOTHING`.

src/python/ensembl/io/genomio/seq_region/rename.py
```diff
@@ -0,0 +1,212 @@
...
+    Args:
+        seq_region: Sequence region.
+
+    Returns:
+        Pairs of source and value for each attribute.
+    """
+
+    attribs = get_attribs(seq_region)
+    attrib_dict = {attrib["source"]: attrib["value"] for attrib in attribs}
+    return attrib_dict
```
JAlvarezJarreta, line 156: `get_attribs()` is only called in `get_attribs_dict()`, so I'm wondering if these two can be combined into a single `get_attributes()` and return the dictionary?

JAlvarezJarreta (approved):
Quite nice new features with the usage of `Enum` I didn't know about :nerd_face:

Dishalodha (approved):
I have tested it, works well. Thank you !

src/python/ensembl/io/genomio/seq_region/rename.py
```diff
@@ -0,0 +1,212 @@
...
+
+    brc_attrib_id = 548
+    if seq_region.operation == Operation.update:
+        stmt = select(SeqRegionAttrib).where(
+            SeqRegionAttrib.seq_region_id == seq_region.seq_region_id,
+            SeqRegionAttrib.attrib_type_id == brc_attrib_id,
+        )
+        seq_attrib = session.scalars(stmt).one()
+        seq_attrib.value = seq_region.brc_name
+    elif seq_region.operation == Operation.insert:
```
Dishalodha, line 183: I am not able to replicate this case if i have such case as 
`if not db_brc_name:
                logging.info(f"Seq region {seqr.name} doesn't have a BRC name")`
it gives the log message however `seq_regions = []` so doesnt apply the insert operation for me

MatBarba: Do you get the message that says the insert operation is not supported (I have only implemented the update part)?

Dishalodha: The value for `return seq_regions` in `get_seq_regions_to_replace` is an empty list for me, so I dont get any message

src/python/ensembl/io/genomio/seq_region/rename.py
```diff
@@ -0,0 +1,209 @@
...
+    elif seq_region.operation == Operation.INSERT:
+        logging.warning(f"Not supported: insertion for {seq_region}")
+    else:
+        logging.warning(f"Cannot update/insert seq_region without clear operation {seq_region}")
+
+
+def rename_seq_regions(dbc: DBConnection, input_map: Path, do_update: bool = False) -> None:
+    """Rename seq_regions in a core from a list of replacements.
+    Args:
+
```
JAlvarezJarreta, line 180:
```suggestion

    Args:
```

JAlvarezJarreta (approved):
Ready to go :+1:

PR #293, opened 2024-02-13
Annotation script to update descriptions

MatBarba:
This script is to be used to compare the descriptions between a core database and a functional_annotation.json file generated by the prepare genome pipeline.

If there are differences, they are reported, and the script can do 2 things:
* Write out the differences (basically the stable_id and the different descriptions)
* Update the core database to replace the descriptions

The current implementation can replace descriptions for genes and transcripts.

This can be thought as a basis for a full fledge functional annotation loader (the things missing are basically the xrefs, the translations and the mobile elements).

### DIFF
--- a/src/python/ensembl/io/genomio/annotation/load.py
+++ b/src/python/ensembl/io/genomio/annotation/load.py
@@ -86,6 +86,8 @@ def load_descriptions(
             "differs": 0,
             "same": 0,
             "no_new": 0,
+            "no_new_xref": 0,
+            "no_new_remove": 0,
         }
         # Compare, only keep the descriptions that have changed
         features_to_update = _get_features_to_update(table, feat_func, feat_data, stats, report, do_update)
@@ -118,11 +120,6 @@ def _get_features_to_update(
 ) -> List[Dict[str, Any]]:
     to_update = []
     for new_feat in feat_func:
-        # Skip if there is no description to transfer
-        if "description" not in new_feat:
-            stats["no_new"] += 1
-            continue
-
         # Check we can find that feature in the core db
         try:
             current_feat = feat_data[new_feat["id"]]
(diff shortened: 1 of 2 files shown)


src/python/ensembl/io/genomio/annotation/load.py
```diff
@@ -0,0 +1,180 @@
...
+        logging.info(f"{len(feat_func)} {table} annotations from {func_file}")
+        feat_data = get_core_data(session, table)
+        logging.info(f"Loaded {len(feat_data)} {table} data")
+
+        stats = {
+            "not_supported": 0,
+            "not_found": 0,
+            "differs": 0,
+            "same": 0,
+            "no_new": 0,
```
JAlvarezJarreta, line 89:
```suggestion
            "not_new": 0,
```
To follow the same nomenclature as the other keys.

src/python/ensembl/io/genomio/annotation/load.py
```diff
@@ -0,0 +1,180 @@
...
+    feat_data: Dict[str, Any],
+    stats: Dict[str, int],
+    report: bool,
+    do_update: bool,
+) -> List[Dict[str, Any]]:
+    to_update = []
+    for new_feat in feat_func:
+        # Skip if there is no description to transfer
+        if "description" not in new_feat:
+            stats["no_new"] += 1
```
JAlvarezJarreta, line 124:
```suggestion
            stats["not_new"] += 1
```

JAlvarezJarreta (approved):
Looks good :+1: 

I would suggest some more consistency (and verbosity) in the variables used in `_get_features_to_update()`, e.g. `current_description` or `current_desc` instead of `cur_dec` (particularly because you have used `current` in other places).

There is also the question of what should happen next if we add the annotation module: the current `gff3` module should be moved inside this one.

JAlvarezJarreta (approved):
Good to go :+1:

PR #294, opened 2024-02-14
Fix db_list when not defined

MatBarba:
Bugfix: the join part fails if no db_list is provided

(part of the dumper pipeline; to reproduce the bug, run the pipeline with no `--db_list` parameter)

### DIFF
--- a/pipelines/nextflow/modules/database/db_factory.nf
+++ b/pipelines/nextflow/modules/database/db_factory.nf
@@ -30,9 +30,13 @@ process DB_FACTORY {
         dbname_re = filter_map.dbname_re ? "--db_regex $filter_map.dbname_re" : ''
 
         // Prepare db_list to write in a file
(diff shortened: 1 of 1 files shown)


JAlvarezJarreta (approved):
Looks good to me, but wouldn't it had been easier just to ensure `db_list` is defined, even if with an empty list? Unless Groovy's `.join()` doesn't work in that case either (taking here the leap of faith from Python).

Dishalodha (approved):

MatBarba:
> Looks good to me, but wouldn't it had been easier just to ensure `db_list` is defined, even if with an empty list? Unless Groovy's `.join()` doesn't work in that case either (taking here the leap of faith from Python).

Yeah that would work, but it requires to do more work for the workflow prep. I think we want to allow this value not to be set.

PR #295, opened 2024-02-14
DBconnection lite

MatBarba:
This a a better version of #242 , to refactor some code and most importantly to add tests (expected coverage 100% for the module).

I've renamed the module from `CoreDatabase` to `DBConnectionLite`, since that's it's goal. The idea here is that we're using `ensembl-py`'s `DBConnection`, but we overwrite the `__init__` to not use `load_all()`. So we only rely on the models from `ensembl.core.models` to get data, which is much faster than allowing `load_all()` (which is not useful and redundant really).

It also contains easy methods to get metadata values from a database.

I suppose this could be added as a suggestion PR to ensembl-py, but I'd rather not touch the `DBConnection` stuff too much...

As a bonus, this PR contains an example of a simple test database fixture built from the SQLalchemy schema (this works here because I only build the `meta` table, which doesn't have MySQL specific dialects, as opposed to `biotype` and `coord_system` which require a `SET`).


src/python/ensembl/io/genomio/database/dbconnection_lite.py
```diff
@@ -0,0 +1,99 @@
...
+    def get_project_release(self) -> str:
+        """Returns the project release number from the database name."""
+
+        match = re.search(_DB_PATTERN_RELEASE, self.db_name)
+        if match:
+            return match.group(1)
+        return ""
+
+
+class DBConnectionLite(DatabaseExtension, DBConnection):
```
JAlvarezJarreta, line 93: I'm curious about the reason behind doing this compared to just renaming `DatabaseExtension` above to `DBConnectionLite` and making it inherit from `DBConnection`, i.e. `class DBConnectionLite(DBConnection):` in line 33.

src/python/tests/database/test_dbconnection_lite.py
```diff
@@ -0,0 +1,116 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Unit testing of `ensembl.io.genomio.database.dbconnection_lite` module.
+"""
+
+from pathlib import Path
+from typing import Optional
+from unittest.mock import patch
+import pytest
+
```
JAlvarezJarreta, line 22:
```suggestion

import pytest
```

src/python/tests/database/test_dbconnection_lite.py
```diff
@@ -0,0 +1,116 @@
...
+from sqlalchemy import create_engine
+from sqlalchemy.engine import URL
+from sqlalchemy.orm import Session
+
+from ensembl.io.genomio.database import DBConnectionLite
+from ensembl.core.models import Base, Meta
+
```
JAlvarezJarreta, line 29:
```suggestion


```
I may be wrong, but it should be 2 spaces between imports and constants based on PEP8.

src/python/tests/database/test_dbconnection_lite.py
```diff
@@ -0,0 +1,116 @@
...
+def db_file_test(tmp_dir: Path) -> Path:
+    """Get a path to a db file."""
+    test_db_file = tmp_dir / "tmp_sqlite_core.db"
+    return test_db_file
+
+
+@pytest.fixture(name="db_engine", scope="session")
+def db_engine_test(db_file: Path):
+    """Get a SQLalchemy engine to a populated database."""
+    db_url = f"sqlite:///{db_file}"
```
JAlvarezJarreta, line 48: Just for future reference, in the ensembl-py plugin there is already a global variable with the selected server you may want to use for your tests. I know it is not applicable in this case, but worth flagging that up.

src/python/tests/database/test_dbconnection_lite.py
MatBarba: 2 empty lines are only around top level classes and functions, cf https://peps.python.org/pep-0008/#blank-lines

MatBarba (commented):
Add couple answers

src/python/ensembl/io/genomio/database/dbconnection_lite.py
JAlvarezJarreta: Ah interesting! Good to know :nerd_face:

PR #299, opened 2024-02-29
Move set_id_prefix to the stable id allocator

MatBarba:
Move the set_prefix to the stable_id allocator (more logical) and add tests for it.

Also reorganize the `init` to be more readable.


JAlvarezJarreta (approved):

Dishalodha (approved):
Nice!

PR #300, opened 2024-02-29
Mbarba/hack/store gene

MatBarba:
Move the `store_gene_annotations` to the annotation module and add tests.

NB: one thing to note is that we have to add the `sub_features` list attrib to the SeqFeatures as those are no longer created by Biopython 1.81+. This attrib is instead added by `bcbio-gff` (cf [this part of its code](https://github.com/chapmanb/bcbb/blob/9c6d83ee3f0491f647a9ecd5947b13c99b478f26/gff/BCBio/GFF/GFFParser.py#L591-L593))

### DIFF
--- a/src/python/tests/gff3/test_extract_annotation.py
+++ b/src/python/tests/gff3/test_extract_annotation.py
@@ -292,13 +292,14 @@ def test_transfer_descriptions(
 
 @pytest.mark.dependency(depends=["add_feature"])
 @pytest.mark.parametrize(
-    "with_cds, num_genes, num_tr, num_cds",
+    "cds_parts, num_genes, num_tr, num_cds",
     [
-        pytest.param(False, 1, 1, 0, id="Store gene without CDS"),
-        pytest.param(True, 1, 1, 1, id="Store gene with CDS"),
+        pytest.param(0, 1, 1, 0, id="Store gene without CDS"),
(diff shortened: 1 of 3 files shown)


src/python/tests/gff3/test_extract_annotation.py
```diff
@@ -288,3 +288,47 @@ def test_transfer_descriptions(
...
+    one_transcript.sub_features = []
+
+    # Add one exon
+    one_exon = SeqFeature(type="exon", id="exon_A")
+    one_transcript.sub_features.append(one_exon)
+
+    # Add a translation (possibly in parts)
+    if cds_parts > 0:
+        for _ in range(1, cds_parts + 1):
+            one_translation = SeqFeature(type="CDS", id="cds_A")
```
JAlvarezJarreta, line 326:
```suggestion
        for i in range(1, cds_parts + 1):
            one_translation = SeqFeature(type="CDS", id=f"cds_{i}")
```
Just in case we want to differentiate the translations.

JAlvarezJarreta (approved):

src/python/tests/gff3/test_extract_annotation.py
MatBarba: We don't: all CDSs under the same transcript are required to have the same ID (though it doesn't matter for this part of the code)

PR #301, opened 2024-02-29
Add unit test for ensembl.io.genomio.genome_metadata.dump module

JAlvarezJarreta:
Updated all methods (review docstrings, improve readability and reduce code complexity) and added unit tests for all of them.

In the case of `get_genome_metadata()` the rework whilst may not be as efficient as the previous code, I believe it is more readable and allows for further meta table checks in case the data found there does not match the expected format, e.g. is `species` meta key always has a subkey but there is an entry without, a `ValueError` is raised.

Added to `__all__` one of the functions that was missing.

### DIFF
--- a/src/python/ensembl/io/genomio/genome_metadata/dump.py
+++ b/src/python/ensembl/io/genomio/genome_metadata/dump.py
@@ -125,58 +125,57 @@ def filter_genome_meta(gmeta: Dict[str, Any]) -> Dict[str, Any]:
     return gmeta_out
 
 
-def check_assembly_version(gmeta_out: Dict[str, Any]) -> None:
-    """Update the assembly version of the genome metadata provided to use an integer.
-    Get the version from the assembly accession as alternative.
+def check_assembly_version(genome_metadata: Dict[str, Any]) -> None:
+    """Updates the assembly version of the genome metadata provided.
+
+    If `version` meta key is not and integer or it is not available, the assembly accession's version
+    will be used instead.
 
     Args:
-        gmeta (Dict[str, Any]): Nested metadata key values from the core metadata table.
-    """
-    assembly = gmeta_out["assembly"]
-    version = assembly.get("version")
+        genome_metadata: Nested metadata key values from the core metadata table.
 
+    Raises:
+        ValueError: If both `version` and the assembly accession's version are not integers or are missing.
+    """
+    assembly = genome_metadata["assembly"]
+    version = assembly.get("version")
     # Check the version is an integer
     try:
         assembly["version"] = int(version)
     except (ValueError, TypeError) as exc:
         # Get the version from the assembly accession
         accession = assembly["accession"]
-        parts = accession.split(".")
-        if len(parts) == 2 and parts[1].isdigit():
-            version = parts[1]
+        version = accession.partition(".")[2]
+        try:
             assembly["version"] = int(version)
-            logging.info(
-                f'Asm version [v{version}] obtained from: assembly accession ({assembly["accession"]}).'
-            )
-        else:
+        except ValueError:
             raise ValueError(f"Assembly version is not an integer in {assembly}") from exc
+        logging.info(f"Assembly version [v{version}] obtained from assembly accession ({accession}).")
     else:
-        logging.info(f"Located version [v{int(version)}] info from meta data.")
+        logging.info(f'Located version [v{assembly["version"]}] info from meta data.')
 
 
-def check_genebuild_version(metadata: Dict[str, Any]) -> None:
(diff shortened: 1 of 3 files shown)


src/python/tests/genome_metadata/test_dump.py
```diff
@@ -0,0 +1,115 @@
...
+    """Tests the `dump.check_assembly_version()` method.
+
+    Args:
+        genome_metadata: Nested metadata key values from the core metadata table.
+        output: Expected change in the genome metadata dictionary.
+        expectation: Context manager for the expected exception (if any).
+    """
+    with expectation:
+        dump.check_assembly_version(genome_metadata)
+        assert not DeepDiff(genome_metadata, output)
```
MatBarba, line 74: Not important, but it might be easier to only check the version value (less expected output to write), e.g.
```suggestion
        assert genome_metadata["assembly"]["version"] == expected_version
```

src/python/ensembl/io/genomio/genome_metadata/dump.py
```diff
@@ -120,58 +125,58 @@ def filter_genome_meta(gmeta: Dict[str, Any]) -> Dict[str, Any]:
...
-        del metadata["genebuild"]["id"]
+    if "version" not in genebuild:
+        try:
+            genebuild_id = genebuild["id"]
+        except KeyError:
+            # pylint: disable=raise-missing-from
+            raise ValueError("No genebuild version or ID found")
+        genome_metadata["genebuild"]["version"] = str(genebuild_id)
+    # Drop genebuild ID since there is a genebuild version
+    genome_metadata["genebuild"].pop("id", None)
```
MatBarba, line 179: Why `pop()` instead of `del`?

MatBarba (approved):
Looks good, some questions about the code

src/python/tests/genome_metadata/test_dump.py
```diff
@@ -0,0 +1,115 @@
...
+            {"assembly": {"accession": "GCA_00000001.1", "version": 1}},
+            does_not_raise(),
+            id="No version, accession with version",
+        ),
+        param(
+            {"assembly": {"accession": "GCA_00000001"}},
+            {},
+            pytest.raises(ValueError),
+            id="No version, accession without version",
+        ),
```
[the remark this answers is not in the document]
JAlvarezJarreta: Based on the current code, if `version` is available, the accession is not checked (thus the selection of parametrizations I have selected). Do you believe is it still worth adding that case?

src/python/ensembl/io/genomio/genome_metadata/dump.py
```diff
@@ -120,58 +125,58 @@ def filter_genome_meta(gmeta: Dict[str, Any]) -> Dict[str, Any]:
...
-        metadata["genebuild"]["version"] = str(gb_id)
-
-    if "id" in genebuild:
-        del metadata["genebuild"]["id"]
+    if "version" not in genebuild:
+        try:
+            genebuild_id = genebuild["id"]
+        except KeyError:
+            # pylint: disable=raise-missing-from
+            raise ValueError("No genebuild version or ID found")
```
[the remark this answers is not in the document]
JAlvarezJarreta: That was my initial thought, but since the error is actually that no keys are found, my logic was that we should raise from 2 exceptions, not one. However, `from` only accepts one exception, thus why I have opted to ignore them.

src/python/tests/genome_metadata/test_dump.py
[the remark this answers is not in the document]
MatBarba: No need, good enough for me

src/python/ensembl/io/genomio/genome_metadata/dump.py
JAlvarezJarreta: `pop` behaves the same as `del` but it also "discards" the exception when passing `None` as second argument, so no need to check if "id" key exists or not.

MatBarba: That's what I thought, it'd be nice if del could something like that but it might be too low level

JAlvarezJarreta: Based on the documentation I was able to find online, this was the only alternative to avoid wrapping `del` around an `if` or `try` clause.

JAlvarezJarreta:
@MatBarba asked for a re-review due to the addition of updated and unit test for `filter_genome_meta()` (messed up the branching/commits and was unable to push review updates in this PR without pushing this as well).

MatBarba (approved):
Looks good now. At some point we might want to test with test databases, but that part is pretty "simple" so the testing is clear.

PR #302, opened 2024-02-29
Add mocks for pipeline tests

MatBarba:
When some external services are done, some parts of the pipeline may fail (`genome_metadata` and `seq_region` prepare in particular). We'd also prefer not to query those services when we run a mock pipeline.

This PR adds a `--mock_run` option to 2 prepare scripts that query ENA, and use it in `stub` commands for the corresponding nextflow modules. Because the output differs in the end, I've also updated the final md5sum for the seq region file.

### DIFF
--- a/src/python/ensembl/io/genomio/genome_metadata/prepare.py
+++ b/src/python/ensembl/io/genomio/genome_metadata/prepare.py
@@ -184,6 +184,7 @@ def get_taxonomy_from_accession(accession: str, base_api_url: str = DEFAULT_API_
     # Use the GenBank accession without version
     gb_accession = accession.replace("GCF", "GCA").split(".")[0]
     response = requests.get(f"{base_api_url}/{gb_accession}", timeout=60)
+    response.raise_for_status()
     entry = ElementTree.fromstring(response.text)
 
     taxon_node = entry.find(".//TAXON")
@@ -232,6 +233,7 @@ def prepare_genome_metadata(
     output_file: PathLike,
     gff3_file: Optional[PathLike] = None,
     base_api_url: str = DEFAULT_API_URL,
+    mock_run: bool = True,
 ) -> None:
     """Updates the genome metadata JSON file with additional information.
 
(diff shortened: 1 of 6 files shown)


src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -232,6 +233,7 @@ def prepare_genome_metadata(
     output_file: PathLike,
     gff3_file: Optional[PathLike] = None,
     base_api_url: str = DEFAULT_API_URL,
+    mock_run: bool = True,
```
JAlvarezJarreta, line 236:
```suggestion
    mock_run: bool = False,
```
I would set this as `False` by default, since I believe we want to this to be the exception, not the norm.

src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -184,6 +184,7 @@ def get_taxonomy_from_accession(accession: str, base_api_url: str = DEFAULT_API_
     # Use the GenBank accession without version
     gb_accession = accession.replace("GCF", "GCA").split(".")[0]
     response = requests.get(f"{base_api_url}/{gb_accession}", timeout=60)
+    response.raise_for_status()
```
JAlvarezJarreta, line 187: Interesting addition :nerd_face:

src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -243,14 +245,19 @@ def prepare_genome_metadata(
...
     """
     genome_data = get_json(input_file)
     # Amend any missing metadata
     add_provider(genome_data, gff3_file)
     add_assembly_version(genome_data)
     add_genebuild_metadata(genome_data)
-    add_species_metadata(genome_data, base_api_url)
+    if mock_run:
+        if not "taxonomy_id" in genome_data["species"]:
+            genome_data["species"]["taxonomy_id"] = 9999999
```
JAlvarezJarreta, line 258:
```suggestion
        genome_data["species"].setdefault("taxonomy_id", 9999999)
```

JAlvarezJarreta (approved):
A good idea deserves a good implementation :+1:

src/python/ensembl/io/genomio/genome_metadata/prepare.py
MatBarba: Yes, that makes sense

src/python/ensembl/io/genomio/genome_metadata/prepare.py
MatBarba: Nice feature :)

PR #303, opened 2024-02-29
Disha/hack feb/update extract genbank

Dishalodha:
I have made the following changes to the code:

1. Made a separate function for all the methods that generate files
2. Assigned value to required arguments directly in __int__
3. Added type hints and doc strings


src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -55,11 +55,12 @@ class UnsupportedData(Exception):
...
 class GenomeFiles(dict):
-    """Store the representation of the genome files created."""
+    """
+    Store the representation of the genome files created.
+    """
 
-    def __init__(self, out_dir: PathLike = Path.cwd()) -> None:
+    def __init__(self, out_dir: PathLike) -> None:
         super().__init__()
-        out_dir = Path(out_dir)
```
MatBarba, line 62: This `Path()` line is required, to use the out_dir / "name" operations below, because PathLike itself doesn't support it.

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -424,26 +489,27 @@ def _get_codon_table(self, seq) -> Optional[int]:
                 return None
         return None
 
-    def _prepare_location(self, organelle):
+    def _prepare_location(self, organelle) -> str:
```
MatBarba, line 492:
```suggestion
    def _prepare_location(self, organelle: str) -> str:
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -358,9 +413,11 @@ def _uniquify_id(self, gene_id, all_ids):
 
         return new_id
 
-    def _write_seq_region_json(self):
+    def _format_write_seq_json(self) -> None:
+        """
+        Add the sequencfe metadata to seq_json based on ensembl requirements
```
MatBarba, line 418:
```suggestion
        Add the sequence metadata to seq_json based on ensembl requirements
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -399,22 +456,30 @@ def _write_seq_region_json(self):
                     "url": "https://www.ncbi.nlm.nih.gov/genbank",
                 },
             }
-            if not seq_obj["added_sequence"]["assembly_provider"]["name"]:
+            if not seq_obj["added_sequence"]["assembly_provider"]:
```
MatBarba, line 459: Not sure what we're doing here: we define `seq_obj["added_sequence"]` just above, so this check will never fail?

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -90,43 +91,38 @@ class FormattedFilesGenerator:
...
+        Args:
+            out_dir: Location to write the files
+        """
+        if out_dir is None:
+            out_dir = Path.cwd()
+        out_dir = Path(out_dir)
 
-        # Output the gff3 file
-        return self.files
+        return out_dir
```
MatBarba, line 117: It might be simpler to just pass Path(out_dir), as it would provide the current working dir if out_dir is not set, so something like:
```
self.files = GenomeFiles(Path(out_dir))
```
(No need to keep `self.out_dir` as well since it's not used anywhere else)

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -138,16 +134,27 @@ def parse_genbank(self, gb_file):
                 if record.id in organella:
                     record.organelle = organella[record.id]
                 self.seq_records.append(record)
+                if len(self.seq_records) >= 1:
+                    self.format_write_record()
+                else:
+                    logging.warning("No records are found in gb_file")
```
MatBarba, line 140: This is in a loop over records, so this part should be unindented once, or even twice since we don't need to file anymore at this point.

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -138,16 +134,27 @@ def parse_genbank(self, gb_file):
...
+        self._write_fasta_dna()
 
-    def _get_organella(self, gb_file):
+    def _get_organella(self, gb_file: PathLike) -> Dict[str, str]:
         """
         Retrieve the organelle from the genbank file, using the specific GenBank object,
         because SeqIO does not support this field
+
+        Args:
+            gb_file: path to the downloaded genbank file
```
MatBarba, line 157:
```suggestion
            gb_file: path to the genbank file
```
We might have the file on hand already ;)

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -159,34 +166,35 @@ def _get_organella(self, gb_file):
                         organella[accession] = organelle
         return organella
 
-    def _write_fasta_dna(self):
+    def _write_fasta_dna(self) -> None:
+        """
+        Generate dna.fasta file with all the sequences in the record
```
MatBarba, line 171:
```suggestion
        Generate a DNA fasta file with all the sequences in the record
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -198,7 +206,35 @@ def _write_genes_gff(self) -> None:
         if num_duplicates > 0:
             raise GBParseError(f"Some {num_duplicates} IDs are duplicated")
 
+    def _write_genes_gff(self, gff_records: List[SeqFeature]) -> None:
+        """
+        Generate gene_models.gff file with th parsed gff_records
```
MatBarba, line 211:
```suggestion
        Generate gene_models.gff file with the parsed gff_records
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -198,7 +206,35 @@ def _write_genes_gff(self) -> None:
         if num_duplicates > 0:
             raise GBParseError(f"Some {num_duplicates} IDs are duplicated")
 
+    def _write_genes_gff(self, gff_records: List[SeqFeature]) -> None:
```
MatBarba, line 209: I recommend you change the name from `gff_records` to `gff_features` in this function

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -198,7 +206,35 @@ def _write_genes_gff(self) -> None:
...
+        Args:
+            gff_records: List of features extracted from the record
+        """
+        logging.debug(f"Write {len(gff_records)} gene records to {self.files['gene_models']}")
+        with self.files["gene_models"].open("w") as gff_fh:
+            GFF.write(gff_records, gff_fh)
+
+    def _write_pep_fasta(self, peptides: List[str]) -> None:
+        """
+        Generate pep.fasta file with the protein ids and sequence
```
MatBarba, line 222:
```suggestion
        Generate a peptide fasta file with the protein ids and sequence
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -245,6 +281,14 @@ def _parse_record(self, record: SeqRecord) -> Tuple[SeqRecord, List[SeqRecord],
     def _parse_gene_feat(
         self, gene_feat: SeqFeature, gene_name: str
     ) -> Tuple[Dict[str, SeqFeature], List[str], List[SeqFeature]]:
+        """
+        Parse the gene records in the genbank file
+
+        Args:
+            gene_feat: Gene features in the record
+            gene_name: Gene name associated with the gene_feat
```
MatBarba, line 289:
```suggestion
        Parse a gene feature from the genbank file

        Args:
            gene_feat: Gene feature to parse
            gene_name: Gene name associated with the gene feature
```

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -307,6 +351,12 @@ def _parse_gene_feat(
         return new_feats, all_ids, peptides
 
     def _parse_rna_feat(self, rna_feat: SeqFeature) -> Tuple[Dict[str, SeqFeature], List[SeqFeature]]:
+        """
+        Parse the rna records in the genbank file
+
+        Args:
+            gene_feat: list of rna found in the record
```
MatBarba, line 358:
```suggestion
        Parse an RNA feature

        Args:
            gene_feat: list of RNA features found in the record
```

MatBarba (changes requested):
Nice cleaning work :) Just a few things to fix and a code simplification suggestion. The rest is just some doc suggestions, feel free to ignore those if you think they don't had much.

Also make sure to run mypy/pylint to find some of the things I've flagged (e.g. the `Path()` thing); this will ensure CICD passes as well

src/python/ensembl/io/genomio/genbank/extract_data.py
Dishalodha: oh thats why mypy was failing, thanks

src/python/ensembl/io/genomio/genbank/extract_data.py
Dishalodha: yes exactly, maybe I will just remove it and add this to the pytest

MatBarba (approved):
This should fix the current problem with synonyms

PR #305, opened 2024-03-04
Disha/update func anno

Dishalodha:
This PR is to add xrefs to functional_annotation.json file as the current code only adds synonyms and uses default dbname (BRC4_Community_Annotation) to load all the xrefs which is incorrect.
I have added an xrefs array for transcripts and translations using provider_name as dbname and name. 

Vasily and I also found a bug in loading load_fann.pl file where the external_dbs were loaded as dictionary and not an array of dict causing the same db_name to overwrite even though the pattern is different in the config file (config/external_db_map/default.txt)


vsitnik (approved):
Thank you!

JAlvarezJarreta (changes requested):
Some minor changes to avoid introducing unintended changes or empty spaces where they should not be.

MatBarba (approved):

src/python/ensembl/io/genomio/gff3/extract_annotation.py
```diff
@@ -21,7 +21,7 @@
     "AnnotationError",
     "FunctionalAnnotations",
 ]
-
+import logging
```
JAlvarezJarreta, line 24:
```suggestion

import logging
```

JAlvarezJarreta (approved):

PR #307, opened 2024-03-05
Fix ENA API

MatBarba:
Add a quick check for the response (to die if something went wrong), and fix the url used for the ENA API: double `//` do not work.


Dishalodha (approved):
Nice 👌

JAlvarezJarreta (approved):
Good catch :+1:

PR #310, opened 2024-03-06
Format outside record loop

MatBarba:
I assume this went ok because we've only loaded single extra sequences until now.

We should add some testing with more than 1 record at some point.


Dishalodha (approved):
yes this should work, I will merge it to mine and test, thank you :)

PR #311, opened 2024-03-06
Mbarba/lcampbell hack

MatBarba:
Misc fixes to make mypy and pylint happy.

I made one significant change to md5_files() to differentiate giving a file name (that should reside in the `dl_dir`), and an actual file path.

The only thing left is the calls in the calls `.assert_called_once*` for which mypy still complains.

### DIFF
--- a/src/python/tests/assembly/test_download.py
+++ b/src/python/tests/assembly/test_download.py
@@ -76,7 +76,7 @@ def test_ftp_connection(
 
     def side_eff_conn(url: str):
         if not url:
-            raise Exception()
+            raise FTPConnectionError()
 
     mock_ftp.connect.side_effect = side_eff_conn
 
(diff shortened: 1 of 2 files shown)


src/python/tests/assembly/test_download.py
```diff
@@ -76,7 +76,7 @@ def test_ftp_connection(
 
     def side_eff_conn(url: str):
         if not url:
-            raise Exception()
+            raise FTPConnectionError()
```
ens-LCampbell, line 79: Cheers for this catch

ens-LCampbell (approved):
Thanks Matthieu you changes all look good to me !

MatBarba:
I've added a type ignore for mypy to not bother checking the assert_called_once from the mocked functions. Since we're not testing ftp itself that should be fine.

PR #312, opened 2024-03-07
Add unit test for ensembl.io.genomio.genome_metadata.prepare module

JAlvarezJarreta:
Updated all methods (review docstrings, improve readability and reduce code complexity) and added unit tests for all of them.


src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -122,9 +122,9 @@ def add_assembly_version(genome_data: Dict) -> None:
     assembly = genome_data["assembly"]
     if not "version" in assembly:
         accession = assembly["accession"]
-        values = accession.split(".")
-        if (len(values) == 2) and values[1]:
-            assembly["version"] = int(values[1])
+        version = accession.partition(".")[2]
+        if version:
+            assembly["version"] = int(version)
```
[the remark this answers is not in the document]
JAlvarezJarreta: `accession.partition(".")` works slight different, as it always returns a tuple with 3 values. But if `.` (or the version number afterwards) is not present, the last 2 values will be `""`. Thus, this check is to prevent a `ValueError` raised when attempting to convert `""` to integer.

src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -75,89 +75,87 @@ class MetadataError(Exception):
...
 
 
 def add_genebuild_metadata(genome_data: Dict) -> None:
```
ens-LCampbell, line 128: Perhaps use a bit more descriptive function name here. Perhaps: `add_genebuild_version_metadata`

src/python/tests/genome_metadata/test_prepare.py
```diff
@@ -0,0 +1,344 @@
...
+        prepare.add_provider(genome_metadata, gff3_file)
+        for section, metadata in output.items():
+            for key, value in metadata.items():
+                assert genome_metadata[section].get(key, None) == value
+
+
+@pytest.mark.parametrize(
+    "genome_file, output",
+    [
+        ("genbank_genome.json", 2),
```
ens-LCampbell, line 108: Not entirely needed but  param set test ID here to conform to other tests. Also just tidies up the result table.

ens-LCampbell (approved):
Minor things only, happy to approve as is. Up to you on comment suggestions

src/python/tests/genome_metadata/test_prepare.py
JAlvarezJarreta: My usual approach is to only add `id` when the print from pytest is going to look cluttered or too complicated to follow. In this case it felt pretty clean and clear (`test_prepare.py::test_add_assembly_version[genbank_genome.json-2]`), but happy to add an `id` if you feel it would be more readable :+1:

src/python/ensembl/io/genomio/genome_metadata/prepare.py
JAlvarezJarreta: This function is adding both `version` and `start_date`, so I think in this case the function name is appropriate. That being said, maybe referring to annotation rather than genebuild would be more accurate, but that can be discussed separaterly.

PR #318, opened 2024-03-13
Pytest is too verbose

MatBarba:
Now that we have more than a hundred tests, the `--verbose` option floods our terminal every time we run pytest.

This PR is to just remove this param by default: now we'll see 1 line per test file, instead of 1 line per individual test.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -179,7 +179,6 @@ warn_unused_configs = true
 addopts = [
     "--import-mode=importlib",
     "--tb=native",
-    "--verbose",
     "--server=sqlite:///",
 ]
 testpaths = ["src/python/tests"]


JAlvarezJarreta (approved):
Should we make this a environment var as we have done with the Python version? This way we can take advantage of the current CI/CD framework in case we want a verbose output (although not sure if this is too useful).

MatBarba:
> Should we make this a environment var as we have done with the Python version? This way we can take advantage of the current CI/CD framework in case we want a verbose output (although not sure if this is too useful).

That would not be helpful for CICD either I think: if tests fail their traceback will be displayed either way. We don't need the details of the tests that pass.

PR #319, opened 2024-03-13
Plugin with version to not check/update when testing

MatBarba:
For the sake of ensuring we can run the pipelines tests offline, we need to fix the nextflow plugin versions in the test config (cf [nextflow doc](https://www.nextflow.io/docs/latest/plugins.html#using-plugins)). That way nextflow doesn't try to connect outside and get the latest version.

With this the current tests (pipelines and pytest) can run completely offline, assuming all dependencies/plugins are installed.


JAlvarezJarreta (approved):
Good catch :+1: The version may not be needed anymore (or at some point soon), since they are planning to rename the v2 version (that was breaking things) to `nf-schemas`. The main reason behind this decision is that v2 is not backwards compatible with v1.

PR #320, opened 2024-03-13
Add manual Nextflow test stage to CI/CD

JAlvarezJarreta:
I have added in the Python pipeline since there is a strong dependency between these 2 (and our current testing system relies on `pytest`).


JAlvarezJarreta:
Need to produce an image with both Python and Nextflow since neither are a service, and cannot use multiple images at once. Will revisit at a later stage when the repository is more stable.

PR #325, opened 2024-03-18
Reverse default in CLI for ID generation

MatBarba:
The default of GFFSimplifier is to make IDs if the ones it finds are invalid/missing. The gff3_process script however defaults to the opposite, so this PR reverses the default for the gff3_process script with `--fail_missing_stable_ids`.


JAlvarezJarreta (approved):
I had the feeling this had been resolved in one the previous PRs... But it was clearly not, good catch!

PR #326, opened 2024-03-19
Prepare genome metadata with datasets

MatBarba:
Instead of using a request to the ENA API, get the metadata from the JSON report from NCBI's datasets.
This simplifies the code (and the tests) since we don't need to check/mock that part.

Now we need to download the dataset file before running the script, so I've updated the Nextflow module to do so.

I've also modified the prepare part in to incorporate the "annotation" part of the genome json if there are annotations, and we can use this information to determine how many files we expect at the end of the Nextflow pipeline (3 or 6).
Also note that I've replaced `concat` by `mix`, since `concat` actually waits for the whole list of each channel before passing the next one, and mix push they as they arrive.

This change makes the `groupTuple()` work correctly: now instead of waiting for all the channels to finish, it starts as soon as it has all the files for a genome.

### DIFF
--- a/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
+++ b/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
@@ -26,9 +26,12 @@ process DATASETS_METADATA {
         
     shell:
         '''
-        datasets summary genome accession !{accession} | jq '.' > ncbi_meta.json
-        if [ ! -s "ncbi_meta.json" ]; then
-            echo "No Metadata from datasets for !{accession}"
+        datasets summary genome accession !{accession} > ncbi_meta.json
+        if [ "$?" -ne 0 ]; then
+            echo "Invalid or unsupported assembly accession: !{accession}"
+            exit 1
+        elif [[ $(jq -r '.total_count' ncbi_meta.json) -eq 0 ]]; then
+            echo "No metadata returned for !{accession}"
             exit 1
         fi
         '''
--- a/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
+++ b/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
@@ -27,6 +27,10 @@ process DATASETS_METADATA {
         '''
         accession=$(jq -r '.["assembly"]["accession"]' !{input_json})
         datasets summary genome accession $accession | jq '.' > ncbi_meta.json
+        if [ ! -s "ncbi_meta.json" ]; then
+            echo "No Metadata from datasets for $accession"
+            exit 1
+        fi
         '''
     
     stub:
--- a/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
+++ b/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
@@ -15,6 +15,7 @@
 
 process DATASETS_METADATA {
     label 'local'
+    label 'cached'
 
     input:
         path(input_json, stageAs: "input_genome.json")
--- a/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
+++ b/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
@@ -13,18 +13,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-process PREPARE_GENOME_METADATA {
+process DATASETS_METADATA {
     label 'local'
 
     input:
-        tuple path(input_json), path(ncbi_json)
+        path(input_json, stageAs: "input_genome.json")
 
     output:
-        path ("genome.json"), emit: genomic_dataset
+        tuple path("input_genome.json", includeInputs: true), path("ncbi_meta.json")
         
     shell:
-    output_json = "genome.json"
-    '''
-    genome_metadata_prepare --input_file !{input_json} --output_file !{output_json} --ncbi_meta !{ncbi_json}
-    '''
+        '''
+        accession=$(jq -r '.["assembly"]["accession"]' !{input_json})
+        datasets summary genome accession $accession | jq '.' > ncbi_meta.json
+        '''
+    
+    stub:
+        """
+        echo '{"reports":[{"organism":{"tax_id":1000}}]}' > ncbi_meta.json
+        """
 }
--- a/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
+++ b/pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
@@ -14,21 +14,21 @@
 // limitations under the License.
 
 process DATASETS_METADATA {
+    tag "$accession"
     label 'local'
     label 'cached'
 
     input:
-        path(input_json, stageAs: "input_genome.json")
+        val(accession)
 
     output:
-        tuple path("input_genome.json", includeInputs: true), path("ncbi_meta.json")
+        tuple val(accession), path("ncbi_meta.json")
         
     shell:
         '''
-        accession=$(jq -r '.["assembly"]["accession"]' !{input_json})
-        datasets summary genome accession $accession | jq '.' > ncbi_meta.json
+        datasets summary genome accession !{accession} | jq '.' > ncbi_meta.json
         if [ ! -s "ncbi_meta.json" ]; then
-            echo "No Metadata from datasets for $accession"
+            echo "No Metadata from datasets for !{accession}"
             exit 1
         fi
         '''
--- a/pipelines/nextflow/modules/genome_metadata/prepare_genome_metadata.nf
+++ b/pipelines/nextflow/modules/genome_metadata/prepare_genome_metadata.nf
@@ -25,7 +25,7 @@ process PREPARE_GENOME_METADATA {
     shell:
     output_json = "genome.json"
     '''
-    accession=$(jq '.["assembly"]["accession"]' genome.json | sed 's/"//g')
+    accession=$(jq '.["assembly"]["accession"]' !{input_json} | sed 's/"//g')
     datasets summary genome accession $accession | jq '.' > ncbi_meta.json
     genome_metadata_prepare --input_file !{input_json} --output_file !{output_json} --ncbi_meta ncbi_meta.json
     '''
--- a/pipelines/nextflow/modules/genome_metadata/prepare_genome_metadata.nf
+++ b/pipelines/nextflow/modules/genome_metadata/prepare_genome_metadata.nf
@@ -25,12 +25,15 @@ process PREPARE_GENOME_METADATA {
     shell:
     output_json = "genome.json"
     '''
-    genome_metadata_prepare --input_file !{input_json} --output_file !{output_json}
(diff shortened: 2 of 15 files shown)


JAlvarezJarreta (commented):
My main concern about this approach is that it forces `datasets` to be installed for it to work (or procure a container with python, ensembl-genomio and nextflow). Since this is out of scope for the time being, what about splitting `prepare_genome_metadata.nf` in two: one where `datasets` is run (so we can use Lahcen's container), and another where `genome_metadata_prepare` is run, so the local environment is used.

MatBarba:
> My main concern about this approach is that it forces `datasets` to be installed for it to work (or procure a container with python, ensembl-genomio and nextflow). Since this is out of scope for the time being, what about splitting `prepare_genome_metadata.nf` in two: one where `datasets` is run (so we can use Lahcen's container), and another where `genome_metadata_prepare` is run, so the local environment is used.

Yes I was considering doing this as well since it'd make caching easier too

MatBarba:
I've separated the datasets download from the preparation step (and I've made it so that the datasets json files are cached, like the downloaded files)

MatBarba:
I have fixed a few issues, notably where I was not in fact using the correct annotation flag. I've also added a separate step so that the download of the summary file only requires an accession value.

pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
```diff
@@ -0,0 +1,40 @@
...
+    output:
+        tuple val(accession), path("ncbi_meta.json")
+        
+    shell:
+        '''
+        datasets summary genome accession !{accession} | jq '.' > ncbi_meta.json
+        if [ ! -s "ncbi_meta.json" ]; then
+            echo "No Metadata from datasets for !{accession}"
+            exit 1
+        fi
```
JAlvarezJarreta, line 33: There are actually 2 types of "errors" from `datasets`: badly formatted accession (prints message in stderr) or accession not matching anything in NCBI (`{"total_counts": 0}`). I believe we want to consider both:
```suggestion
        datasets summary genome accession !{accession} > ncbi_meta.json
        if [ "$?" -ne 0 ]; then
            echo "Invalid or unsupported assembly accession: !{accession}"
            exit 1
        elif [[ $(jq -r '.total_count' ncbi_meta.json) -eq 0 ]]; then
            echo "No metadata returned for !{accession}"
            exit 1
        fi
```
Note that I have removed `| jq '.'` because if left it "nullifies" `$?` (makes it 0 whatever `datasets` returned).

pipelines/nextflow/modules/genome_metadata/prepare_genome_metadata.nf
```diff
@@ -14,23 +14,18 @@
...
 
     output:
-        path ("genome.json"), emit: genomic_dataset
+        tuple val(accession), path("genome.json"), emit: genomic_dataset
         
     shell:
     output_json = "genome.json"
     '''
-    genome_metadata_prepare --input_file !{input_json} --output_file !{output_json}
+    genome_metadata_prepare --input_file !{input_json} --output_file !{output_json} --ncbi_meta !{ncbi_json}
```
JAlvarezJarreta, line 29: Why have you removed the stub?

pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
```diff
@@ -0,0 +1,40 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+process DATASETS_METADATA {
+    tag "$accession"
+    label 'local'
+    label 'cached'
+
```
JAlvarezJarreta, line 20: Shall we add the container here? Since `NXF_SINGULARITY_CACHEDIR` is pointing to were our current container is, it should work for us. We can wait until the next containers meeting to see how we release it to the world after that. I assume `nextflow.config` will need to be also updated to enable singularity.

pipelines/nextflow/workflows/genome_prepare/main.nf
```diff
@@ -49,22 +51,31 @@ def meta_from_genome_json(json_path) {
         prod_name = data.species.production_name
         publish_dir = prod_name
     }
+    has_annotation = false
+    if (data.annotation) {
+        has_annotation = true
+    }
 
     return [
+        // id: prod_name,
```
JAlvarezJarreta, line 60:
```suggestion
```

src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -75,15 +68,15 @@ class MetadataError(Exception):
...
+def add_provider(genome_metadata: Dict, ncbi_data: Dict) -> None:
     """Updates the genome metadata adding provider information for assembly and gene models.
 
     Assembly provider metadata will only be added if it is missing, i.e. neither `"provider_name"` or
     `"provider_url"` are present. The gene model metadata will only be added if `gff3_file` is provided.
 
     Args:
         genome_data: Genome information of assembly, accession and annotation.
-        gff3_file: Path to GFF3 file to use as annotation source for this genome.
+        ncbi_data: Report data from datasets for this accession.
```
JAlvarezJarreta, line 79:
```suggestion
        ncbi_data: Report data from NCBI datasets.
```

src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -141,95 +134,35 @@ def add_genebuild_metadata(genome_data: Dict) -> None:
...
-    If `"taxonomy_id"` is already present in the species metadata, nothing is added. The `"taxonomy_id"`,
-    `"scientific_name"` and `"strain"` will be fetched from the taxonomy information linked to the given
-    accession.
+def add_species_metadata(genome_metadata: Dict, ncbi_data: Dict) -> None:
+    """Adds tax_id and name from the NCBI dataset report.
 
     Args:
         genome_metadata: Genome information of assembly, accession and annotation.
-        base_api_url: Base API URL to fetch the taxonomy data from.
+        ncbi_data: Report data from datasets for this accession.
```
JAlvarezJarreta, line 142:
```suggestion
        ncbi_data: Report data from NCBI datasets.
```

src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -239,20 +172,19 @@ def prepare_genome_metadata(
     Args:
         input_file: Path to JSON file with genome metadata.
         output_file: Output directory where to generate the final `genome.json` file.
-        gff3_file: Path to GFF3 file to use as annotation source for this genome.
-        base_api_url: Base API URL to fetch the taxonomy data from.
-        mock_run: Do not call external taxonomy service.
+        ncbi_meta: JSON file from NCBI datasets for the accession.
```
JAlvarezJarreta, line 175:
```suggestion
        ncbi_meta: JSON file from NCBI datasets.
```

src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -264,19 +196,13 @@ def main() -> None:
     parser.add_argument_dst_path(
         "--output_file", required=True, help="Output path for the new genome metadata file"
     )
-    parser.add_argument_src_path("--gff3_file", help="GFF3 file to use as annotation source")
-    parser.add_argument(
-        "--base_api_url", default=DEFAULT_API_URL, help="API URL to fetch the taxonomy data from"
+    parser.add_argument_src_path(
+        "--ncbi_meta", required=True, help="JSON file from NCBI datasets for this accession."
```
JAlvarezJarreta, line 200: I have removed the reference to `accession` in previous documentation because there is no such argument anywhere.
```suggestion
        "--ncbi_meta", required=True, help="JSON file from NCBI datasets for this genome."
```

src/python/tests/genome_metadata/test_prepare.py
```diff
@@ -90,13 +83,13 @@ def test_add_provider(
     Args:
         json_data: JSON test file parsing fixture.
         genome_file: Genome metadata JSON file.
-        gff3_file: GFF3 file.
+        ncbi_data: Report from datasets.
```
JAlvarezJarreta, line 86:
```suggestion
        ncbi_data: Report from NCBI datasets.
```

src/python/tests/genome_metadata/test_prepare.py
```diff
@@ -150,195 +143,93 @@ def test_add_genebuild_metadata(
...
     """Tests the `prepare.prepare_genome_metadata()` method.
 
     Args:
-        mock_*: A mock of `ensembl.io.genomio.genome_metadata.prepare.*` functions.
-        tmp_path: Test's unique temporary directory fixture.
-        data_dir: Module's test data directory fixture.
-        assert_files: File diff assertion fixture.
+        input_filename: Genome json input.
+        ncbi_filename: NCBI dataset json input.
+        expected_filename: Genome json input expected output.
```
JAlvarezJarreta, line 225:
```suggestion
        input_filename: Input genome JSON file.
        ncbi_filename: NCBI dataset report JSON file.
        expected_filename: Expected output genome JSON file.
```

JAlvarezJarreta (changes requested):
Quite a nice update of the codebase, showing the potential of `datasets` CLI tool :nerd_face:

pipelines/nextflow/modules/genome_metadata/prepare_genome_metadata.nf
MatBarba: This part no longer needs to be mocked: the ncbi_meta generated by the datasets module already produces the mock file in its own stub

pipelines/nextflow/modules/genome_metadata/datasets_metadata.nf
MatBarba: When we agree on the container yes

src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -141,95 +134,35 @@ def add_genebuild_metadata(genome_data: Dict) -> None:
...
 
 
-def add_species_metadata(genome_metadata: Dict, base_api_url: str = DEFAULT_API_URL) -> None:
-    """Adds missing species metadata from its taxonomy based on the genome's accession.
-
-    If `"taxonomy_id"` is already present in the species metadata, nothing is added. The `"taxonomy_id"`,
-    `"scientific_name"` and `"strain"` will be fetched from the taxonomy information linked to the given
-    accession.
+def add_species_metadata(genome_metadata: Dict, ncbi_data: Dict) -> None:
+    """Adds tax_id and name from the NCBI dataset report.
```
JAlvarezJarreta, line 138:
```suggestion
    """Adds taxonomy ID, scientific name and strain (if present) from the NCBI dataset report.
```

JAlvarezJarreta (approved):
One tiny docstring update, but this is now ready to go!

PR #328, opened 2024-03-20
Add xref test, fix small bug

MatBarba:
Add test for get_xrefs and fix a bug when the DB xref value contains a colon


Dishalodha (approved):
Looks good, could we make a PR against main and try to merge it sooner?

PR #331, opened 2024-03-21
fixing BioSample typo and adding more RR/MVP related meta_key duplicates

vsitnik:
tested on metazoam load of `rs8t_mgen/metadata`

### DIFF
--- a/scripts/gff_metaparser/genmetaconf/metaconf.py
+++ b/scripts/gff_metaparser/genmetaconf/metaconf.py
@@ -122,9 +122,11 @@ class MetaConf:
             if "strain" in qualifiers:
                 strain = qualifiers["strain"][0]
                 self.update("species.strain", strain)
+                self.update("strain.type", "strain")
             elif "isolate" in qualifiers:
                 strain = qualifiers["isolate"][0]
                 self.update("species.strain", strain)
+                self.update("strain.type", "isolate")
             if "db_xref" in qualifiers:
                 taxon_id_pre = list(filter(lambda x: x.startswith("taxon:"), qualifiers["db_xref"]))[0]
                 if taxon_id_pre:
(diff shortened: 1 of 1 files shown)


scripts/gff_metaparser/genmetaconf/metaconf.py
```diff
@@ -244,6 +253,18 @@ def update_derived_data(self, defaults=None, update_annotation_related=False):
             syns.append((sci_name_words[0][0] + sci_name_words[1][:5]).lower())
         if syns and self.add_generated_species_aliases:
             self.update("species.alias", syns)
+        # picking assembly.alt_accession
+        if not self.get("assembly.alt_accession"):
+            asm_acc_insdc = self.get("assembly.accession_insdc")
+            asm_acc_refseq = self.get("assembly.accession_refseq")
+            if asm_acc_refseq and asm_acc_insdc:
+                if "refseq" in _ann_source.lower() or \
+                       "refseq" in self.get("annotation.provider_url").lower():
```
MatBarba, line 262: What does this check do? I would think the annotation does not matter for the `assembly.*accession*` meta keys

MatBarba (approved):
Looks ok, just one question about a check

scripts/gff_metaparser/genmetaconf/metaconf.py
vsitnik: if there's /refseq/i in url

vsitnik: We use this logic only for annotations from RefSeq annotation.
In theory, we can have genbank (insdc) assembly, that have a corresponding (GCF) RefSeq annotation, but a non-refseq annotation was loaded

MatBarba: I believe we will only set the accession_refseq if the genome is from RefSeq so that check should not be necessary. It's fine to keep this although you need to make sure the annotation.source of provider_name are set properly

vsitnik: thank you. keeping, though removing url checking bit

PR #332, opened 2024-03-21
GFF3 simplifier refactoring and tests

MatBarba:
Lots of refactoring in the simplifier module with it's company of tests.


src/python/tests/gff3/test_records.py
```diff
@@ -0,0 +1,73 @@
...
+    data_dir: Path, in_gff: PathLike, excluded: Optional[List[str]], expected_loaded: List[str]
+) -> None:
+    """Test loading GFF records from file."""
+    input_gff = data_dir / in_gff
+
+    records = Records()
+    if excluded is None:
+        records.from_gff(input_gff)
+    else:
+        records.from_gff(input_gff, excluded)
```
JAlvarezJarreta, line 46:
```suggestion
    records.from_gff(input_gff, excluded)
```
I'm always unsure if we should also "test" default values, but at the same time we want the test to fail if the type changes, e.g. is excluded does not support `None` anymore, it should cause our tests to fail so they are updated accordingly.

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+
+from Bio.SeqFeature import SeqFeature
+import pytest
+from pytest import param, raises
+
+from ensembl.io.genomio.gff3.exceptions import GFFParserError
+from ensembl.io.genomio.gff3.simplifier import GFFSimplifier
+
+
+def check_one_feature(input_gff: PathLike, output_gff: PathLike, check_function: str) -> SeqFeature:
```
JAlvarezJarreta, line 30: Could this be made a fixture instead?

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+    [
+        param("ok_gene.gff", "ok_gene.gff", does_not_raise(), id="ok gene"),
+        param("lone/transcript.gff", "lone/transcript_simped.gff", does_not_raise(), id="lone transcript"),
+        param("lone/trna.gff", "lone/trna_simped.gff", does_not_raise(), id="lone tRNA"),
+        param("lone/rrna.gff", "lone/rrna_simped.gff", does_not_raise(), id="lone rRNA"),
+    ],
+)
+def test_create_gene_for_lone_transcript(
+    data_dir: Path,
+    tmp_dir: Path,
```
JAlvarezJarreta, line 58:
```suggestion
    tmp_path: Path,
    data_dir: Path,
```
I really need to find out the root of this `tmp_dir` fixture and remove it. Note I have also swapped the order because it is usually more convenient to put "external" fixtures first.

Apply the same in all the other tests below.

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+    assert_files: Callable,
+    in_gff: PathLike,
+    expected_gff: PathLike,
+    expectation: ContextManager,
+) -> None:
+    """Test gene create gene for lone transcript."""
+    input_gff = data_dir / in_gff
+    output_gff = tmp_dir / Path(in_gff).name
```
JAlvarezJarreta, line 66:
```suggestion
    output_gff = tmp_path / Path(in_gff).name
```

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+    tmp_dir: Path,
+    assert_files: Callable,
+    in_gff: PathLike,
+    expected_gff: PathLike,
+    expectation: ContextManager,
+) -> None:
+    """Test gene create gene for lone CDS."""
+    input_gff = data_dir / in_gff
+    output_gff = tmp_dir / Path(in_gff).name
+    with expectation:
```
JAlvarezJarreta, line 92: Same as above.

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+    tmp_dir: Path,
+    assert_files: Callable,
+    in_gff: PathLike,
+    expected_gff: PathLike,
+    expectation: ContextManager,
+) -> None:
+    """Test gene create gene for lone transcript."""
+    input_gff = data_dir / in_gff
+    output_gff = tmp_dir / Path(in_gff).name
+    with expectation:
```
JAlvarezJarreta, line 67: The code doesn't raise exception, so I would suggest to remove this argument and the additional context manager.

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+        param("C_gene_segment", "immunoglobulin", "IG_C_gene", does_not_raise(), id="C immunoglobulin"),
+        param("C_gene_segment", "ig", "IG_C_gene", does_not_raise(), id="C ig"),
+        param("V_gene_segment", "t-cell", "TR_V_gene", does_not_raise(), id="V t-cell"),
+        param("V_gene_segment", "T_cell", "TR_V_gene", does_not_raise(), id="V T_cell"),
+        param("V_gene_segment", "Lorem Ipsum", "", raises(GFFParserError), id="V T_cell"),
+    ],
+)
+def test_format_gene_segments(
+    in_type: str,
+    name: Optional[List[str]],
```
JAlvarezJarreta, line 190:
```suggestion
    name: str,
```
This change may allow you to put all the arguments in a single line.

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+        param("ok_gene.gff", "ok_gene.gff", False, does_not_raise(), id="ok type, fail"),
+    ],
+)
+def test_simpler_gff3_skip(
+    data_dir: Path,
+    tmp_dir: Path,
+    assert_files: Callable,
+    in_gff: PathLike,
+    expected_gff: PathLike,
+    skip_unrecognized: Optional[bool],
```
JAlvarezJarreta, line 305:
```suggestion
    skip_unrecognized: bool,
```
Signature changed before.

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+        simp = GFFSimplifier()
+        simp.simpler_gff3(input_gff)
+        simp.records.to_gff(output_gff)
+        assert_files(output_gff, data_dir / expected_gff)
+
+
+@pytest.mark.parametrize(
+    "in_gff, expected_gff, skip_unrecognized, expectation",
+    [
+        param("bad_gene_type.gff", "", None, raises(GFFParserError), id="Unset skip unrecognized, fail"),
```
JAlvarezJarreta, line 287:
```suggestion
        param("bad_gene_type.gff", "", False, raises(GFFParserError), id="Unset skip unrecognized, fail"),
```

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+    skip_unrecognized: Optional[bool],
+    expectation: ContextManager,
+) -> None:
+    """Test simplifying genes from GFF3 files."""
+    input_gff = data_dir / in_gff
+    output_gff = tmp_dir / in_gff
+    with expectation:
+        simp = GFFSimplifier()
+        if skip_unrecognized is not None:
+            simp.skip_unrecognized = skip_unrecognized
```
JAlvarezJarreta, line 314:
```suggestion
        simp = GFFSimplifier(skip_unrecognized=skip_unrecognized)
```

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+        simp.simpler_gff3(input_gff)
+        simp.records.to_gff(output_gff)
+        assert_files(output_gff, data_dir / expected_gff)
```
JAlvarezJarreta, line 317: I would leave only inside this context block the elements excepted to raise the exception (if any), to avoid other code raising it and making the test pass when it shouldn't.

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,365 @@
...
+    ],
+)
+def test_gffsimplifier_with_genome(
+    data_dir: Path,
+    tmp_dir: Path,
+    assert_files: Callable,
+    genome_file: Optional[PathLike],
+    in_gff: PathLike,
+    expected_gff: PathLike,
+    expectation: ContextManager,
```
JAlvarezJarreta, line 353: Same as before with no expected exceptions.

JAlvarezJarreta (changes requested):
Quite a significant chunk of updates, but nicely done :+1:

src/python/ensembl/io/genomio/gff3/simplifier.py
```diff
@@ -108,90 +113,172 @@ def simpler_gff3(self, in_gff_path: PathLike) -> None:
...
+        new_gene.qualifiers["ID"] = new_gene.id
+        feat.id = self.stable_ids.generate_transcript_id(new_gene.id, 1)
+        feat.qualifiers["ID"] = feat.id
+
+        return new_gene
+
+    def create_gene_for_lone_cds(self, feat: SeqFeature) -> SeqFeature:
+        """Returns a gene created for a lone CDS."""
+        if feat.type != "CDS":
+            return feat
```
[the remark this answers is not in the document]
MatBarba: I see one big reason to keep the if in the function: it ensures that we check the `SeqFeature` in the place where we use it. The alternative is to have the `if` outside, in which case there is no guarantee that the `SeqFeature` passed is what we expect. Or we would need to add another check in the function too, which would make a lot redundant code.

Also: pylint complains when we have too much conditions in the same place...

src/python/ensembl/io/genomio/gff3/simplifier.py
```diff
@@ -38,26 +37,32 @@
...
     """List of GFF3 SeqRecords."""
 
-    def to_gff(self, out_gff_path: PathLike) -> None:
-        """Print out the current list of records in a GFF3 file.
+    def from_gff(self, in_gff_path: PathLike, excluded: Optional[List[str]] = None) -> None:
+        """Load records from a GFF3 file."""
+        if excluded is None:
+            excluded = []
+        with Path(in_gff_path).open("r") as in_gff_fh:
+            for record in GFF.parse(in_gff_fh):
```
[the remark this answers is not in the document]
MatBarba: The `Records` object is general enough to be used for that already

MatBarba (commented):
Thanks, lots of good points

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,457 @@
...
+    data_dir: Path,
+    tmp_dir: Path,
+    assert_files: Callable,
+    in_gff: PathLike,
+    expected_gff: PathLike,
+    expectation: ContextManager,
+) -> None:
+    """Test normalizing miRNA genes."""
+    input_gff = data_dir / in_gff
+    output_gff = tmp_dir / Path(in_gff).name
```
JAlvarezJarreta, line 452: Why not making `in_gff` a `Path` directly?

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,457 @@
...
+    with expectation:
+        simp = GFFSimplifier()
```
JAlvarezJarreta, line 454: Same as before.

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,457 @@
...
+        simp.simpler_gff3(input_gff)
+        simp.records.to_gff(output_gff)
+        assert_files(output_gff, data_dir / expected_gff)
```
JAlvarezJarreta, line 457: Could this go outside the `with` block as you have done above?

JAlvarezJarreta (changes requested):
Quite the work! But nicely done :+1:

src/python/tests/gff3/test_simplifier.py
```diff
@@ -0,0 +1,457 @@
...
+    assert_files: Callable,
+    in_gff: PathLike,
+    expected_gff: PathLike,
+) -> None:
+    """Test gene create gene for lone transcript."""
+    input_gff = data_dir / in_gff
+    output_gff = tmp_path / Path(in_gff).name
+    new_feat = check_one_feature(input_gff, output_gff, "create_gene_for_lone_transcript")
+    if new_feat:
+        assert_files(output_gff, Path(data_dir / expected_gff))
```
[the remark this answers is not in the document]
MatBarba: You're right, the underlying culprit `simpler_gff3_feature()` should always return a `SeqFeature`, and raise an exception otherwise. `Now check_one_feature()` do not return anything anymore so I've removed the `if new_feat` parts.

src/python/ensembl/io/genomio/gff3/simplifier.py
```diff
@@ -108,94 +122,179 @@ def simpler_gff3(self, in_gff_path: PathLike) -> None:
...
+
+    def simpler_gff3_feature(self, gene: SeqFeature) -> SeqFeature:
+        """Creates a simpler version of a GFF3 feature.
 
-        with Path(in_gff_path).open("r") as in_gff_fh:
-            for record in GFF.parse(in_gff_fh):
-                if record.id in self.exclude_seq_regions:
-                    logging.debug(f"Skip seq_region {record.id}")
-                    continue
+        Raises InvalidGFFFeature if the feature is invalid/skippable.
```
JAlvarezJarreta, line 149:
```suggestion
        Raises:
            IgnoredFeatureError: If the feature type is ignored.
            UnsupportedFeatureError: If the feature type is not supported.
```

JAlvarezJarreta (approved):
One tiny docstring update as we are done! :muscle:

PR #333, opened 2024-03-22
Revert temporary pytest version limit as pytest-workflow has now been fixed

JAlvarezJarreta:


MatBarba (approved):
No problem running things with this

PR #336, opened 2024-03-27
CICD when scripts are modified

MatBarba:
Black is run on all folders, but CICD is not run when scripts/ Python scripts are modified. This lead to issues later on.


vsitnik (approved):
ok :)

PR #337, opened 2024-03-27
Fix integrity error report

MatBarba:
The current integrity has a major bug where the integrity errors found are not reported (only the files parsing errors are).

Explanation: there are 2 lists of errors: manifest.errors (from parsing the files) and integrity.errors (from comparing the data). Only the first one raised an exception, the second one was not raised or shown in the logs.

This PR ensure both are raised (and the manifest issues are raised first).

Also fix the way errors are stored in integrity.errors.


Dishalodha (approved):
Tested, works !

PR #338, opened 2024-03-27
Fix seq compare in integrity

MatBarba:
2 fixes:
- Use the synonyms from the seq_region json in case the GFF uses synonyms
- The number of only_feat seq regions was incorrect

### DIFF
--- a/src/python/ensembl/io/genomio/manifest/check_integrity.py
+++ b/src/python/ensembl/io/genomio/manifest/check_integrity.py
@@ -154,7 +154,7 @@ class Manifest:
             if readable_hash != md5sum:
                 raise InvalidIntegrityError(f"Invalid md5 checksum for {file_path}")
 
-    def prepare_integrity_data(self) -> None:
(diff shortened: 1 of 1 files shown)


src/python/ensembl/io/genomio/manifest/check_integrity.py
```diff
@@ -154,7 +154,7 @@ def _check_md5sum(self, file_path: Path, md5sum: str) -> None:
             if readable_hash != md5sum:
                 raise InvalidIntegrityError(f"Invalid md5 checksum for {file_path}")
 
-    def prepare_integrity_data(self) -> None:
+    def prepare_integrity_data(self) -> None:  # pylint: disable=too-many-branches
```
JAlvarezJarreta, line 157: :scream:

MatBarba: Yeah, I'd rather have the fix available first then we can think about making pylint happy (or change the number of branches we allow after all)

Dishalodha (approved):
Tested !

PR #345, opened 2024-04-15
RefSeq gene IDs update

MatBarba:
Change how we handle RefSeq gene IDs:
* The ID is valid if it starts with LOC*
* Otherwise we fallback to GeneID_*
* If there is no Dbxref GeneID, we generate a new ID

Rational: the current ID validity check is too broad for RefSeq so we sometimes we end up with some gene names as IDs. Restricting to LOC* and falling back directly to GeneIDs will avoid any such gene names to be used a IDs.

### DIFF
--- a/src/python/ensembl/io/genomio/gff3/id_allocator.py
+++ b/src/python/ensembl/io/genomio/gff3/id_allocator.py
@@ -19,7 +19,7 @@ __all__ = ["StableIDAllocator", "InvalidStableID"]
 from dataclasses import dataclass, field
 import logging
 import re
-from typing import Dict, List, Set
+from typing import Dict, List, Optional, Set
 
 from Bio.SeqFeature import SeqFeature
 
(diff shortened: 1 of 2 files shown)


src/python/ensembl/io/genomio/gff3/id_allocator.py
```diff
@@ -173,32 +173,40 @@ def normalize_gene_id(self, gene: SeqFeature) -> str:
...
+                (db, value) = xref.split(":")
+                if db != "GeneID":
+                    continue
+                new_gene_id_base = f"{db}_{value}"
+                new_gene_id = new_gene_id_base
+                number = 1
+                while new_gene_id in self._loaded_ids:
+                    number += 1
+                    new_gene_id = f"{new_gene_id_base}_{number}"
+                    if number > 10:
```
Dishalodha, line 201: why do we raise a error only when the number exceeds 10, why not when we find more than 1 unique id?

MatBarba: Some genomes use the same GeneID for several different genes in the same annotation, so this is to allow some room for those.

src/python/ensembl/io/genomio/gff3/id_allocator.py
```diff
@@ -173,32 +173,40 @@ def normalize_gene_id(self, gene: SeqFeature) -> str:
         prefixes = ["gene-", "gene:"]
         new_gene_id = StableIDAllocator.remove_prefix(gene.id, prefixes)
 
+        is_valid = False
+        # Special case for RefSeq: only valid Gene IDs are LOC*
+        if refseq:
+            if new_gene_id.startswith("LOC"):
+                is_valid = True
+        else:
+            is_valid = self.is_valid(new_gene_id)
```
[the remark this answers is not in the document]
MatBarba: Any suggestion? I think it's clear enough for now

PR #347, opened 2024-04-16
Fix after merge

MatBarba:
Somehow logging module was removed but was necessary. Also fix a test.


src/python/ensembl/io/genomio/gff3/extract_annotation.py
```diff
@@ -23,6 +23,7 @@
 ]
 
 from os import PathLike
+import logging
```
JAlvarezJarreta, line 26: Wasn't this removed because `pylint` flagged it as unused?

JAlvarezJarreta (approved):
Happy with the additions.

PR #348, opened 2024-04-16
Store RefSeq gene xref

MatBarba:
Immediately store the RefSeq gene ID as xref if it's ID is in the LOC* format.

(NB: the other RefSeq features transcripts and translations are already stored because they are in the dbxref)

### DIFF
--- a/src/python/ensembl/io/genomio/gff3/extract_annotation.py
+++ b/src/python/ensembl/io/genomio/gff3/extract_annotation.py
@@ -96,7 +96,7 @@ class FunctionalAnnotations:
         # Add RefSeq ID xref if it looks like one
         if self.provider_name == "RefSeq":
             if feature.type == "gene" and feature.id.startswith("LOC"):
-                xref_dbs = [x["dbname"] for x in all_xref]
+                xref_dbs = {x["dbname"] for x in all_xref}
                 if "RefSeq" not in xref_dbs:
                     all_xref.append({"dbname": "RefSeq", "id": feature.id})
         return all_xref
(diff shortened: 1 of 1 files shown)


src/python/ensembl/io/genomio/gff3/extract_annotation.py
```diff
@@ -78,27 +78,26 @@ def __init__(self, provider_name: str = "") -> None:
...
-            xrefs = {"dbname": dbname, "id": name}
-            all_xref.append(xrefs)
+                xrefs = {"dbname": dbname, "id": name}
+                all_xref.append(xrefs)
 
         # Add RefSeq ID xref if it looks like one
         if self.provider_name == "RefSeq":
             if feature.type == "gene" and feature.id.startswith("LOC"):
                 xref_dbs = {x["dbname"] for x in all_xref}
                 if "RefSeq" not in xref_dbs:
```
Dishalodha, line 98: Does this condition mean we cannot have more than 1 RefSeq xref? 
eg in such cases
```python
{
     	"description": "uncharacterized LOC129721096, transcript variant X4",
        "id": "LOC129721096_t1",
        "object_type": "transcript",
        "xrefs": [
            {
             	"dbname": "GeneID",
                "id": "129721096"
            },
	    {
                "dbname": "RefSeq",
                "id": "XR_008727343.1"
     	    }
        ]
    },

```

MatBarba: Yes, there should never be several RefSeq IDs for the same feature. In this example it's not a gene so it's not impacted; we store correctly the transcript RefSeq ID.

Dishalodha: ahh yes this is a transcript, makes sense. Thank you!!

Dishalodha (approved):

PR #349, opened 2024-04-16
Mbarba/load using xref

MatBarba:
In case the stable ID does not match (possible when e.g. stable IDs have been allocated), this PR allows matching the features with their xrefs, assuming the ID from the GFF can match an xref ID

Also contains an important fix to correctly match the transcript IDs in the query


Dishalodha (approved):
I am happy with this :)

PR #352, opened 2024-04-22
Bug fix for assembly tracker

ens-LCampbell:
Due to ncbi updating their version of datasets tool since assembly tracker was first introduced, meant when running assembly tracker caused it to crash out as a result of the returned query type test failing (expected _string_) but got a _list_. 

The result contained the JSONl assembly query in addition to:
Announcement message reporting a newer version of datasets was available. 

Fix checks result type returned and removes user announcement messages when using non-latest versions of datasets.


JAlvarezJarreta (approved):
Good catch :+1:

PR #354, opened 2024-04-23
Set GitLab CI/CD runner tag to ensure the correct job environment

JAlvarezJarreta:
It seems some available shared runners do now have Docker, which was causing failure of some of our jobs. Setting a default job tack to request all of them to have docker should address this problem.


vsitnik (approved):
cool. thanks!

PR #356, opened 2024-04-26
Move archive utils to ensembl-py

JAlvarezJarreta:
I find these extremely useful for many teams, so it is best to put them somewhere where they can be seen and used. Once https://github.com/Ensembl/ensembl-py/pull/48 is approved and merged, the CI/CD will pass here and it can be merged.

Not sure if it is worth doing the same for the JSON utils we have. Happy to move them as well if you see value on that.


Dishalodha (approved):
yes, these seem like more generic functions and I don't see a issue with moving it to ensembl-py.

PR #359, opened 2024-05-09
Minor updates of our codebase

JAlvarezJarreta:
Some variable renames to avoid clashing with classes, file name harmonisation and adding a missing `__init__.py` file.


ens-LCampbell (approved):
Looks good to me. Assuming its been tested and works as expected :)

PR #361, opened 2024-05-13
Update documentation logo

JAlvarezJarreta:
Ensembl is getting a new logo and it may be this the right time to start updating it.


ens-LCampbell (approved):
Minor edits indeed, all good. I assume the image is fine however I can't see it 'unable to render rich display'

PR #365, opened 2024-05-13
Remove outdated project layout dir tree

ens-LCampbell:
Quick change...
- Remove doc tree of src directory files as its becomes out of sync with codebase.


JAlvarezJarreta (approved):
I believe this removal makes sense :+1:

PR #366, opened 2024-05-16
Allow mapping mRNA IDs from the Apollo FASTA file to protein IDs

MatBarba:
The FASTA file that we get for the patch build from Apollo usually uses the mRNA IDs for the protein sequences. Instead of asking everytime to chagne the FASTA file to use the protein IDs instead, I've modified our loader to allow matching protein IDs (already loaded in the core database) using the mRNA/transcript IDs from the Apollo FASTA file.

Additionally, I've removed the log "Not in file" which can contain a lot of sequences (since the patch build FASTA file only contains the genes that have changed). This prevents that whole list of sequences to fill the email at the end of the loading.

### DIFF
--- a/src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/ReportSeqEdits.pm
+++ b/src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/ReportSeqEdits.pm
@@ -212,7 +212,7 @@ sub protein_seq_report {
 
         if (exists $protein{$tn_id}) {
           $feat_id = $tn_id;
-        }elsif (exists $protein{$tt_id}) {
+        } elsif (exists $protein{$tt_id}) {
           $feat_id = $tt_id;
         }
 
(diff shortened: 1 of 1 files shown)


src/perl/Bio/EnsEMBL/Pipeline/Runnable/EG/LoadGFF3/ReportSeqEdits.pm
```diff
@@ -207,11 +208,19 @@ sub protein_seq_report {
       if ($translation) {
         my $tn_id  = $translation->stable_id;
         my $db_seq = $translation->seq;
+        my $feat_id = "";
 
         if (exists $protein{$tn_id}) {
-          $translations{$tn_id} = 1;
+          $feat_id = $tn_id;
+        }elsif (exists $protein{$tt_id}) {
```
JAlvarezJarreta, line 215:
```suggestion
        } elsif (exists $protein{$tt_id}) {
```

JAlvarezJarreta (approved):
Looks good, particularly the reduction of verbosity on the log :+1: Made a tiny suggestion just because :grin:

PR #367, opened 2024-05-16
Keep transcript ID if the gene ID is conserved on a gene with one transcript

MatBarba:
The gene diff script can keep the gene ID if they overlap enough, but the transcript IDs are systematically regenerated in this case. If a gene has a single transcript and is updated, it makes sense to also keep the transcript ID for that gene.

This PR keeps the transcript ID if a gene only has one transcript before and after the patch build. It also transfers the translation ID if any (which is always paired with the transcript).

I've also increased the resources of the Nextflow module transfer_metadata since it tends to require more resources and it can be run in a separate thread. I've had to add a fix to ensure all the operations are finished before checking the patch.

### DIFF
--- a/pipelines/nextflow/modules/patch_build/transfer_metadata.nf
+++ b/pipelines/nextflow/modules/patch_build/transfer_metadata.nf
@@ -14,7 +14,8 @@
 // limitations under the License.
 
 process TRANSFER_METADATA {
-    label 'local'
+    memory "8 GB"
+    time "2h"
 
     input:
         path(waited_file)
--- a/scripts/brc4/patch_build/transfer_ids.pl
+++ b/scripts/brc4/patch_build/transfer_ids.pl
@@ -58,10 +58,11 @@ $reg_path = $opt{new_registry};
 $registry->load_all($reg_path);
 
 # Transfer the IDs to the new genes using the mapping
-my $missed_transcripts = transfer_gene_ids($registry, $species, $mapping, $old_genes, $opt{update});
+my ($missed_trs, $mapped_trs) = transfer_gene_ids($registry, $species, $mapping, $old_genes, $opt{update});
 
 # Print out the missed transcripts
-print_missed($missed_transcripts, $opt{out_transcripts});
+print_list($missed_trs, $opt{missed_transcripts});
+print_list($mapped_trs, $opt{mapped_transcripts});
 
 ###############################################################################
 sub load_mapping {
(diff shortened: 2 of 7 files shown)


scripts/brc4/patch_build/transfer_ids.pl
```diff
@@ -161,64 +168,84 @@ sub transfer_gene_ids {
...
-    update_translation_id($tra, $old_translation, $translation) if $update;
-    $stats->{translation}++;
+sub transfer_transcript_id {
+  my ($old_transcript, $transcript, $stats, $tra, $update) = @_;
+  $logger->debug("Update id for " . $transcript->stable_id . " to " . $old_transcript->{"stable_id"});
+  
+  my $translation = $transcript->translation;
+  my $old_translation_id = $old_transcript->{"translation_id"};
+
+  # Stop if we don't have translation in both old and new transcript
```
JAlvarezJarreta, line 219:
```suggestion
  # Stop if we don't have translation in both old and new transcripts
```

scripts/brc4/patch_build/transfer_ids.pl
```diff
@@ -242,7 +269,8 @@ sub usage {
     --new_registry <path> : Ensembl registry
     --species <str>       : production_name of one species
     --mapping <path>      : Old gene ids to new gene ids mapping file
-    --out_transcripts <path: List of transcript IDs not replaced by an older ID
+    --missed_transcripts <path: List of transcript IDs not replaced by an older ID
+    --mapped_transcripts <path: List of transcript IDs replaced by an older ID
```
JAlvarezJarreta, line 273:
```suggestion
    --missed_transcripts <path> : List of transcript IDs not replaced by an older ID
    --mapped_transcripts <path> : List of transcript IDs replaced by an older ID
```

JAlvarezJarreta (approved):
Couple of minor suggestions, but the changes proposed look good and the code has improved its readability. Well done :+1:

pipelines/nextflow/modules/patch_build/transfer_metadata.nf
```diff
@@ -14,7 +14,8 @@
 // limitations under the License.
 
 process TRANSFER_METADATA {
-    label 'local'
+    memory "8 GB"
+    time "2h"
```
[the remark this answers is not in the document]
MatBarba: Potentially but I don't plan to optimize this pipeline too much given that we're not sure how long we'll do patch buidls

PR #368, opened 2024-05-17
External db map fix

vsitnik:


config/external_db_map/default.txt
```diff
@@ -133,11 +139,16 @@ GenBank	GenBank_translation	translation
...
 GenBank_translation	GenBank_translation	translation
 
+HAMAP	HAMAP	translation
+
 INTERPRO	Interpro	translation
 
 MEROPS	MEROPS	translation
 
 PDB	PDB	translation
+PFAM	PFAM	translation
```
ens-LCampbell, line 149: Just wondering why such mapping only added now ? Would of expected PFAM in place

config/external_db_map/default.txt
```diff
@@ -88,6 +91,9 @@ GenBank	GenBank_transcript	transcript
 GB	GenBank_transcript	transcript
 GenBank_transcript	GenBank_transcript	transcript
 
+JGIDB	JGI_TRANSCRIPT	transcript
```
ens-LCampbell, line 94: Thanks !! This DB was the only failure for the 8 new genome loads Im currently running

ens-LCampbell (approved):
Seems ok to me, thanks again

config/external_db_map/default.txt
vsitnik: PFAM is in place, but we just allowing it to be attached only to the peptide/CDS/translation xrefs

PR #369, opened 2024-05-21
mRNA without gene

MatBarba:
- Add explicit support for mRNA (with exon and CDS) without a gene parent.
- Also change the IDs in lone transcript cases: keep the ID on the transcript (and their children) and generate a new ID for the gene, namely the transcript ID + "_gene"
- Also change the gene type to pseudogene if the CDS has a `pseudo=true` attribute
- Update the tests accordingly

### DIFF
--- a/src/python/ensembl/io/genomio/gff3/simplifier.py
+++ b/src/python/ensembl/io/genomio/gff3/simplifier.py
@@ -197,7 +197,7 @@ class GFFSimplifier:
         return self.clean_gene(gene)
 
     def create_gene_for_lone_transcript(self, feat: SeqFeature) -> SeqFeature:
-        """Returns a gene for lone transcripts: 'gene' for tRNA/rRNA, and 'ncRNA_gene' for all others.
+        """Returns a gene for lone transcripts: 'gene' for tRNA/rRNA/mRNA, and 'ncRNA_gene' for all others.
 
         Args:
             feat: The transcript for which we want to create a gene.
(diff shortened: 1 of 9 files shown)


src/python/ensembl/io/genomio/gff3/simplifier.py
```diff
@@ -207,18 +207,28 @@ def create_gene_for_lone_transcript(self, feat: SeqFeature) -> SeqFeature:
...
+        if feat.type == "mRNA":
+            is_pseudo = False
+            for subfeat in feat.sub_features:
+                pseudo_qual = subfeat.qualifiers.get("pseudo", [""])[0]
+                if subfeat.type == "CDS" and pseudo_qual == "true":
+                    is_pseudo = True
+                    del subfeat.qualifiers["pseudo"]
+                    break
+            if is_pseudo:
+                new_gene.type = "pseudogene"
```
JAlvarezJarreta, line 231: What about the two lines that have been removed for `tRNA` and `rRNA`?
```
        feat.id = self.stable_ids.generate_transcript_id(new_gene.id, 1)
        feat.qualifiers["ID"] = feat.id
```

MatBarba: Should I keep transfering the transcript ID to the gene for tRNAs and rRNA, but not mRNAs? Or transfer it for all?

JAlvarezJarreta (approved):
Happy if CI/CD is happy. Weird bugs are always weird, no matter when.

PR #370, opened 2024-05-21
Gene segment info from the CDS

MatBarba:
We get the gene segment information from the transcript itself, but in some cases the information is in their CDS, this PR adds this extra check.

Rewrite the code with a new exception `GeneSegmentError` when we can't get the information to make the proper biotype.

### DIFF
--- a/src/python/ensembl/io/genomio/gff3/simplifier.py
+++ b/src/python/ensembl/io/genomio/gff3/simplifier.py
@@ -414,24 +414,31 @@ class GFFSimplifier:
             transcript: Gene segment transcript feature.
 
         Raises:
-            GFFParserError: Missing or unexpected transcript's standard name.
+            GeneSegmentError: Unable to get the segment type information from the feature.
         """
         if transcript.type not in ("C_gene_segment", "V_gene_segment"):
             return transcript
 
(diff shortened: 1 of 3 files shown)


src/python/ensembl/io/genomio/gff3/simplifier.py
```diff
@@ -414,25 +414,43 @@ def format_gene_segments(self, transcript: SeqFeature) -> SeqFeature:
...
+            seg_type = self._get_segment_type(cdss[0])
+
+        # Change V/C_gene_segment into a its corresponding transcript names
+        transcript.type = f"{seg_type}_{transcript.type.replace('_segment', '')}"
         return transcript
 
+    def _get_segment_type(self, feature: SeqFeature) -> str:
+        """Infer if a segment is "IG" (immunoglobulin) of "TR" (t-cell) from the feature attribs.
+
+        Raises: GeneSegmentError if the attribs are missing that information.
```
JAlvarezJarreta, line 439:
```suggestion
        Raises:
            GeneSegmentError: If the feature attribs are missing that information.
```

JAlvarezJarreta (approved):
Just a tiny suggestion, but looks good to me :+1:

JAlvarezJarreta (approved):

PR #371, opened 2024-05-22
Add no_fail option to integrity for nextflow to not die

MatBarba:
Add a `--no_fail` argument to the `check_integrity.py` script. The default behaviour is still to fail when an integrity fails (most notably for the genome loader).

This PR changes the default manifest integrity Nextflow module to use that `--no_fail` option so as to not kill the whole pipeline when an integrity check fails. Instead, `integrity.out` files are created in each folder and only kept if there have been errors. .

Impacted pipelines: prepare_genome, additional_seq_prepare, dump_files

PS: We should think about writing a final report (sent by email or printed to screen/file) to show what went wrong at the end of a pipeline.

### DIFF
--- a/pipelines/nextflow/modules/manifest/integrity.nf
+++ b/pipelines/nextflow/modules/manifest/integrity.nf
@@ -23,11 +23,21 @@ process CHECK_INTEGRITY {
         tuple val(meta), path(manifest_files)
     
     output:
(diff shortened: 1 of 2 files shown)


vsitnik (approved):
Cool. Thank you!

PR #372, opened 2024-05-22
Mbarba/match synonym

MatBarba:
We're matching functional_annotation ID to the core stable_id, and we've recently also added a match from the xrefs.
This PR adds a match to the synonyms as well (this improves the matching to transfer descriptions and xrefs from an older core).

* Also make the match case insensitive
* Rewrite the code to make it a bit clearer

### DIFF
--- a/src/python/ensembl/io/genomio/annotation/load.py
+++ b/src/python/ensembl/io/genomio/annotation/load.py
@@ -16,7 +16,7 @@
 
 import logging
 from pathlib import Path
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Tuple, Union
 
 from sqlalchemy.orm import Session
 from sqlalchemy import and_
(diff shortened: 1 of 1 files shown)


src/python/ensembl/io/genomio/annotation/load.py
```diff
@@ -139,10 +141,37 @@ def load_descriptions(
                 session.commit()
 
 
+def _get_cur_feat(
+    feat_data: Dict[str, FeatStruct], new_feat: Dict[str, Any], match_xrefs: bool = False
+) -> Union[FeatStruct, None]:
```
JAlvarezJarreta, line 146:
```suggestion
) -> Optional[FeatStruct]:
```

JAlvarezJarreta (approved):
The refactor looks sensible. Just a tiny change but good otherwise.

PR #375, opened 2024-05-22
Incorporate use of datasets-cli into genome prepare pipeline

ens-LCampbell:
**Aim:**:
 Make use of container with datasets-cli tool to pull both metadata and data file from NCBI. 

Changes:
- Addition of bespoke nextflow module (download_asm_with_datasets) process to work with datasets-cli, retained old module process (download_asm_data) for now. 
- Update nextflow config to make use of container in process with use of _withlabel: 'datasets_cli'_
- Various updates to genome_metadata/extend and seq_region/prepare modules to align with change in assembly report meta keys
- Added the latest (ok it was, now is _semi latest_ 16.17.1) definition file for datasets-cli SIF creation
- Update the test suite of genome_prepare pipeline, including data differences and MD5 checksums

@MatBarba I have tested this over, if you can think of anything else let me know thanks 
- inputs with and without annotation
- with and without brc_mode enabled
- using '-stub'


pipelines/nextflow/modules/download/download_asm_with_datasets.nf
```diff
@@ -0,0 +1,63 @@
...
+            emit: opt_set, optional: true
+
+    shell:
+        '''
+        datasets download genome accession !{meta.accession} \
+        --include seq-report,genome,gbff,gff3,protein \
+        --assembly-source all \
+        --dehydrated \
+        --filename !{meta.accession}_ncbi.zip \
+        --no-progressbar
```
JAlvarezJarreta, line 45:
```suggestion
            --include seq-report,genome,gbff,gff3,protein \
            --assembly-source all \
            --dehydrated \
            --filename !{meta.accession}_ncbi.zip \
            --no-progressbar
```
Just a suggestion as I find the extra indentation more readable for commands, but happy to leave it as it is.

pipelines/nextflow/modules/download/download_asm_with_datasets.nf
```diff
@@ -0,0 +1,63 @@
...
+
+        unzip !{meta.accession}_ncbi.zip -d !{meta.accession}_ncbi
+
+        datasets rehydrate --directory !{meta.accession}_ncbi \
+        --gzip \
+        --no-progressbar
```
JAlvarezJarreta, line 51:
```suggestion
            --gzip \
            --no-progressbar
```

pipelines/nextflow/subworkflows/genome_prepare/main.nf
```diff
@@ -18,7 +18,7 @@
 
 // Import modules/subworkflows
 include { CHECK_JSON_SCHEMA as CHECK_JSON_SCHEMA_GENOME } from '../../modules/schema/check_json_schema.nf'
-include { DOWNLOAD_ASM_DATA } from '../../modules/download/download_asm_data.nf'
+include { DOWNLOAD_ASM_NCBI_DATSETS } from '../../modules/download/download_asm_with_datasets.nf'
```
JAlvarezJarreta, line 21:
```suggestion
include { DOWNLOAD_ASM_NCBI_DATASETS } from '../../modules/download/download_asm_with_datasets.nf'
```

pipelines/nextflow/subworkflows/genome_prepare/main.nf
```diff
@@ -45,9 +45,9 @@ workflow GENOME_PREPARE {
         checked_genome = CHECK_JSON_SCHEMA_GENOME(genomic_dataset).verified_json
 
         // Download genome data files. Files may or may not include gene models (GFF3) and/or peptides.
-        DOWNLOAD_ASM_DATA(checked_genome)
-        download_min = DOWNLOAD_ASM_DATA.out.min_set
-        download_opt = DOWNLOAD_ASM_DATA.out.opt_set
+        DOWNLOAD_ASM_NCBI_DATSETS(checked_genome)
+        download_min = DOWNLOAD_ASM_NCBI_DATSETS.out.min_set
+        download_opt = DOWNLOAD_ASM_NCBI_DATSETS.out.opt_set
```
JAlvarezJarreta, line 50:
```suggestion
        DOWNLOAD_ASM_NCBI_DATASETS(checked_genome)
        download_min = DOWNLOAD_ASM_NCBI_DATASETS.out.min_set
        download_opt = DOWNLOAD_ASM_NCBI_DATASETS.out.opt_set
```

pipelines/nextflow/workflows/nextflow.config
```diff
@@ -114,4 +114,7 @@ process {
         storeDir = { "${params.cache_dir}/${task.process.tokenize(':')[-1].toLowerCase()}/${task.tag}" }
         afterScript = { "sleep ${params.storeDir_latency}" }
     }
+    withLabel: 'datasets_cli' {
+        container = 'library://lcampbell/ensembl-genomio/ncbi-datasets-v16.17.1:latest'
```
JAlvarezJarreta, line 118: Something we definitely need to replace soon with a proper URL for the sake of usability by other people.

JAlvarezJarreta (changes requested):
Great improvement to GenomIO :+1: Good to see `datasets` is in good shape to be used for production.

Since v16.17.3 is already out, I'm suggesting we use the latest to avoid having to do another update at a later stage (the definition file will need to be renamed as well).

pipelines/nextflow/workflows/nextflow.config
ens-LCampbell: Sounds good, until then we stick with this hack I guess :)

JAlvarezJarreta (approved):

pipelines/nextflow/tests/workflows/test_genome_prepare.yml
```diff
@@ -47,5 +47,5 @@
     # Manifest depends on the genome checksum, so also date dependent
     - path: ./tmp/genome_prepare/GCA_017607445.1/manifest.json
     - path: ./tmp/genome_prepare/GCA_017607445.1/seq_region.json
-      md5sum: 28518b0c7cbc19a2890a6b347367a82f
+      md5sum: 6a45dc461c53e33dde33807c6def7b63
```
MatBarba, line 50: The content of the file has significantly changed. Here's a diff:
```
        "synonyms": [                                                   "synonyms": [
            {                                                               {
                "name": "CM029948.1",                                           "name": "CM029948.1",
                "source": "GenBank"                                             "source": "GenBank"
            },                                                              },
            {                                                               {
                "name": "Mitochondrion",                      |                 "name": "MT",
                "source": "INSDC"                                               "source": "INSDC"
            },                                                              },
            {                                                               {
                "name": "",                                   |                 "name": "HcG217B07",
                "source": "INSDC_submitted_name"                                "source": "INSDC_submitted_name"
            },                                                <
            {                                                 <
                "name": "",                                   <
                "source": "RefSeq"                            <
            }                                                               }
```

MatBarba (changes requested):
One important typo, and need to check the output of seq_region.json (the new version includes some empty values, and has seemingly lost another)

MatBarba:
Another general comment: we're generating an assembly report file derived from the jsonl file, but it has different attributes than the usual one that we get from ftp. Then would it not make more sense to parse the json file directly (and leave the current seq_region/extend.py as deprecated/legacy to parse assembly report from ftp)?

ens-LCampbell:
> Another general comment: we're generating an assembly report file derived from the jsonl file, but it has different attributes than the usual one that we get from ftp. Then would it not make more sense to parse the json file directly (and leave the current seq_region/extend.py as deprecated/legacy to parse assembly report from ftp)?

This would make more sense Indeed ! Basically remove the extra step of converting to TSV from seq_region.jsonl. However it does work albeit slightly extra processing overhead, the issue is loss of INSDC_submitted_name in the datasets world. If Im not mistake, this is incoming soonish is that right @JAlvarezJarreta ?

PR #379, opened 2024-05-29
Disha/nextflow notify

Dishalodha:
I have added the notification option in the main config to get an email every time we run a nextflow pipeline. Currently it uses the basic configuration in the email, containing workflow status and the execution summary.

### DIFF
--- a/pipelines/nextflow/workflows/nextflow.config
+++ b/pipelines/nextflow/workflows/nextflow.config
@@ -39,7 +39,7 @@ trace {
     // filter = 'name =~ /foo.*/ && status == "FAILED"'
 }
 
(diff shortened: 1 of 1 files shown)


pipelines/nextflow/workflows/nextflow.config
```diff
@@ -39,6 +39,11 @@ trace {
     // filter = 'name =~ /foo.*/ && status == "FAILED"'
 }
 
+notification{
```
JAlvarezJarreta, line 42:
```suggestion
notification {
```

JAlvarezJarreta (approved):
So simple... Thanks!

PR #384, opened 2024-06-17
Disha/nextflow utils

Dishalodha:
I have create a nextflow utils file, this function can be used in our pipeline to handle multiple parameters.


JAlvarezJarreta (commented):
If this is a more nf-core approach, can you please move the contents of `ensembl-genomio/pipelines/nextflow/modules/utils/utils.nf` here as well? Otherwise, it may be less trouble to move this code there.

JAlvarezJarreta (approved):
Code looks great :+1: Thank you!

JAlvarezJarreta (changes requested):

JAlvarezJarreta (approved):

PR #386, opened 2024-06-19
Disable notification in main nextflow.config

MatBarba:
This should be a per-user configuration, not by default


JAlvarezJarreta (approved):
We may want to add this to the nextflow documentation so people know this functionality is there and how to enable it.

PR #387, opened 2024-06-20
Dump with container

MatBarba:
Use the container defined in https://github.com/MatBarba/ensembl-scripts and available on dockerhub at https://hub.docker.com/repository/docker/matthieubarba/ensembl-scripts/general. In principle this would make it possible to run the pipeline without the whole ensembl environment installed.

This makes it easy to use the dumper scripts defined in that repo, which are simplified versions of the hive runnable dumpers from ensembl-production.

The current nextlfow dumper pipeline can now dump all files that the old hive pipeline could:
- AGP
- FASTA DNA
- FASTA Peptide
- JSON Functional annotation
- JSON genome metadata
- JSON seq_regions
- JSON sequence attribs
- SQL Core dump
- GFF3 gene models

All the dumper scripts and helpers (gt, json_schema...) are available in the container, which contains both the genomio python scripts, and a selection of Perl dumpers not yet (if ever) migrated to Python.


pipelines/nextflow/workflows/nextflow.config
```diff
@@ -39,9 +39,8 @@ trace {
     // filter = 'name =~ /foo.*/ && status == "FAILED"'
 }
 
-notification {
+singularity {
```
JAlvarezJarreta, line 42: We need to remember to make `singularity` a dependency of the Nextflow environment in this case.

JAlvarezJarreta (commented):
Looks good.

Couple of ideas/suggestions:
- merge `dump_fasta_dna.nf` and `dump_fasta_peptides.nf` and just pass the `type` as an input, to simplify the code
- In the future we may want to split the `dump_gff3.nf` script in two, so that way you only need "simple" containers, i.e. one for the script, and one for `gt`

MatBarba:
> Looks good.
> 
> Couple of ideas/suggestions:
> 
>     * merge `dump_fasta_dna.nf` and `dump_fasta_peptides.nf` and just pass the `type` as an input, to simplify the code
I kept it separate in case we want to use different scripts and parameters for both cases (or if we want to do some other specific and different post-processing on the output). E.g, there is a script to dump protein sequences in ensembl-analysis, and we may use that one instead of the simple one I made.

>     * In the future we may want to split the `dump_gff3.nf` script in two, so that way you only need "simple" containers, i.e. one for the script, and one for `gt`
It makes more sense to do the gt tidy and validation in the same process where the GFF3 is created. This way if the validation fails we just need to rerun that process (otherwise only the validation step fails, and the GFF3 dumper step can't be rerun when that's where things actually failed).

JAlvarezJarreta:
@MatBarba yes, those are valid points, so the content looks good as it is in that case :+1:

pipelines/nextflow/workflows/nextflow.config
MatBarba: We need to have either docker or singularity here, otherwise the scripts called need to be in the local environment

ens-LCampbell (approved):

src/python/ensembl/io/genomio/manifest/generate.py
```diff
@@ -73,26 +75,52 @@ def get_files_checksums(self):
...
         return manifest_files
 
+    def _prepare_object_name(self, subfile: Path, name: str, manifest_dict: dict) -> str:
+        # Prepare object name
+        obj_name = "file"
+        try:
+            # If we recognize the suffix, then the name is the part after the last _
+            if subfile.suffix == f".{name}":
+                obj_name = subfile.stem.split(sep="_")[-1]
+            # If we recognize the end of the name, then the name is the part before the last _
```
JAlvarezJarreta, line 108:
```suggestion
            # If we recognize the end of the name, then the name is the part before the last "_"
```

src/python/ensembl/io/genomio/manifest/generate.py
```diff
@@ -73,26 +75,52 @@ def get_files_checksums(self):
...
+            pass
+
+        # Add number if duplicate name
+        obj_name_base = obj_name
+        num = 1
+        while obj_name in manifest_dict:
+            obj_name = f"{obj_name_base}.{num}"
+            num += 1
+            if num >= 10:
+                raise ValueError(f"Too many files with same name {obj_name_base}")
```
JAlvarezJarreta, line 121:
```suggestion
        count = 1
        while obj_name in manifest_dict:
            obj_name = f"{obj_name_base}.{count}"
            count += 1
            if count >= 10:
                raise ValueError(f"Too many files with same name {obj_name_base}")
```

pipelines/nextflow/modules/seq_region/dump_agp.nf
```diff
@@ -0,0 +1,52 @@
...
+        dump_agp.pl \
+            --host $db.server.host \
+            --port $db.server.port \
+            --user $db.server.user \
+            $password_arg \
+            --dbname $db.server.database \
+            -v \
+            --output_dir ./
+        
+        nfiles=\$(ls *.agp | wc -l)
```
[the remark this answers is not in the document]
MatBarba: Indeed, but when passing `ls` to a pipe, the spaces are changed to new lines (=-1 is by default). I'll still commit this just for peace of mind...

src/python/ensembl/io/genomio/manifest/generate.py
```diff
@@ -73,26 +75,51 @@ def get_files_checksums(self):
...
+        except IndexError:
+            pass
+
+        # Add number if duplicate name
+        obj_name_base = obj_name
+        count = 1
+        while obj_name in manifest_dict:
+            obj_name = f"{obj_name_base}.{count}"
+            count += 1
+            if count >= 10:
```
ens-LCampbell, line 119: Any reason for 10 files being the threshold ?

pipelines/nextflow/modules/seq_region/dump_agp.nf
```diff
@@ -0,0 +1,52 @@
...
+            $password_arg \
+            --dbname $db.server.database \
+            -v \
+            --output_dir ./
+        
+        nfiles=\$(ls -1 *.agp | wc -l)
+        if [ "\$nfiles" -eq "0" ]; then
+            touch empty.agp
+        fi
+
```
JAlvarezJarreta, line 44:
```suggestion
```

pipelines/nextflow/workflows/dumper_pipeline/main.nf
```diff
@@ -16,15 +16,21 @@
...
     'stats': 3,
     'fasta_pep': 1,
-    'fasta_dna': 1
+    'fasta_dna': 1,
+    'seq_attrib': 1,
 ]
 default_selection = default_selection_map.keySet() as ArrayList
 params.db_list = ''
+params.server_url = ''
+params.host = ''
```
JAlvarezJarreta, line 33: Shouldn't all the params definitions with their default value be defined in the JSON schema and/or the `nextflow.config` file?

pipelines/nextflow/workflows/dumper_pipeline/main.nf
```diff
@@ -38,16 +44,29 @@ if (params.brc_mode) {
...
+            "port": params.port,
+            "user": params.user,
+        ]
     if (params.password) {
         server["password"] = params.password
     }
+        server.url = generate_url("mysql", params.host, params.port, params.user, params.password)
+    }
+    if (!server.url or !server.host) {
+        log.info paramsHelp("Server URL (--server_url) or parameters needed (--host, --port, --user, --password)")
```
JAlvarezJarreta, line 66: The previous code will do nothing if both URL and host are provided. I believe it may be easier/better to just provide a parameter precedence:
```suggestion
    if (params.server_url) {
        server = extract_url_args(params.server_url)
        server.url = params.server_url
    } else if (params.host) {
        server = [
            "host": params.host,
            "port": params.port,
            "user": params.user,
        ]
        if (params.password) {
            server["password"] = params.password
        }
        server.url = generate_url("mysql", params.host, params.port, params.user, params.password)
    } else {
        log.info paramsHelp(
            "Server URL (--server_url) or parameters needed (--host, --port, --user, --password)"
        )
```

src/python/ensembl/io/genomio/seq_region/dump.py
```diff
@@ -39,8 +40,9 @@
 from ensembl.utils.logging import init_logging_with_args
 
 
-_ROOT_DIR = Path(__file__).parent / "../../../../../.."
-_DEFAULT_MAP = _ROOT_DIR / "config/external_db_map/default.txt"
+default_map_res = files("ensembl.io.genomio.data.external_db_map").joinpath("default.txt")
+with as_file(default_map_res) as default_map_path:
+    _DEFAULT_MAP = default_map_path
```
JAlvarezJarreta, line 45: This is quite handy, isn't it? Although I'm a little bit annoyed that you have to use the `with` now to access the file, when you didn't before.

pipelines/nextflow/workflows/dumper_pipeline/main.nf
MatBarba: Those are only here to ensure we don't get `none` values and raise a warning from the validation, they are not true "default" values.

src/python/ensembl/io/genomio/manifest/generate.py
MatBarba: Just an arbitrary number in case there are a lot of files. If we encounter a database with a lot of those (right now only from AGP) then we can update the number.

src/python/ensembl/io/genomio/seq_region/dump.py
MatBarba: RIght, I think I need to actually test this because I would expect the resource/file to become unavailable after the with context is closed...

MatBarba: I guess that's the price to pay to have access to resources that could be something else than files.

src/python/ensembl/io/genomio/manifest/generate.py
JAlvarezJarreta: Shall we add this number as an argument for the function, with default value set to 10? This way it can be easily changed for specific cases/conditions if needed.

PR #394, opened 2024-07-08
Disha/update nextflow utils

Dishalodha:
Added functions to nextflow utils to create a url and extract arguments from it, here is the ticket[ ENSMETAZOA-122](https://www.ebi.ac.uk/panda/jira/browse/ENSMETAZOA-122). However, here we do expose the password not sure how we could handle it.

### DIFF
--- a/pipelines/nextflow/modules/utils/utils.nf
+++ b/pipelines/nextflow/modules/utils/utils.nf
@@ -51,12 +51,12 @@ def parse_list_param(String multi_value = '', List<String> allowed_values, Boole
     return all_params
 }
 
-def generate_url(protocol, host, port, user, pass=null, database=null) {
+def generate_url(protocol, host, port, user, password=null, database=null) {
     // Generate a URL when all the credentials are provided
     base_url = "${protocol}://${user}@${host}:${port}"
 
-    if (pass) {
-        base_url = "${protocol}://${user}:${pass}@${host}:${port}"
+    if (password) {
+        base_url = "${protocol}://${user}:${password}@${host}:${port}"
     }
     if (database) {
         base_url += "/${database}"
@@ -74,9 +74,9 @@ def extract_url_args(url_string) {
     def url_parts = url_string.tokenize('/@')
     def user_info = url_parts[1].split(":")
     def user = user_info[0]
-    def pass = user_info.length > 1 ? user_info[1] : null
+    def password = user_info.length > 1 ? user_info[1] : null
     def (host, port) = url_parts[2].split(":")
     def database = url_parts.size() > 3 ? url_parts[3] : null
 
-    return [user: user, pass: pass, host: host, port: port, database: database]
+    return [user: user, password: password, host: host, port: port, database: database]
 }
--- a/pipelines/nextflow/modules/utils/utils.nf
+++ b/pipelines/nextflow/modules/utils/utils.nf
@@ -50,3 +50,17 @@ def parse_list_param(String multi_value = '', List<String> allowed_values, Boole
     }
     return all_params
 }
+
+def generate_url(host, port, user, pass=null, dbname=null) {
+    base_url = "${protocol}://${user}@${host}:${port}"
+
+    if (pass) {
+        base_url = "${protocol}://${user}:${password}@${host}:${port}"
+    }
+
+    if (dbname) {
+        base_url += "/${dbname}"
+    }
+    return base_url
+}
+
--- a/pipelines/nextflow/modules/utils/utils.nf
+++ b/pipelines/nextflow/modules/utils/utils.nf
@@ -51,20 +51,22 @@ def parse_list_param(String multi_value = '', List<String> allowed_values, Boole
     return all_params
 }
 
-def generate_url(host, port, user, pass=null, dbname=null) {
+def generate_url(protocol, host, port, user, pass=null, database=null) {
+    // Generate a URL when all the credentials are provided
     base_url = "${protocol}://${user}@${host}:${port}"
 
     if (pass) {
-        base_url = "${protocol}://${user}:${password}@${host}:${port}"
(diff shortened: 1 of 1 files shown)


pipelines/nextflow/modules/utils/utils.nf
```diff
@@ -50,3 +50,38 @@ def parse_list_param(String multi_value = '', List<String> allowed_values, Boole
...
+        base_url = "${protocol}://${user}:${pass}@${host}:${port}"
+    }
+    if (database) {
+        base_url += "/${database}"
+    }
+    return base_url
+}
+
+
+def extractMySQLArguments(urlString) {
```
JAlvarezJarreta, line 68:
```suggestion
def extract_mysql_args(url) {
```
To follow the same approach as in other functions.

pipelines/nextflow/modules/utils/utils.nf
```diff
@@ -50,3 +50,38 @@ def parse_list_param(String multi_value = '', List<String> allowed_values, Boole
...
+    def user_info = urlparts[0].split(":")
+    def user = user_info[0]
+    def pass = user_info.length > 1 ? user_info[1]:null
+
+    def host_database = urlparts[1].split("/")
+    def host_info = host_database[0].split(":")
+    def host = host_info[0]
+    def port = host_info[1]
+
+    def database = host_database.length > 1 ? host_database[1] : null
```
JAlvarezJarreta, line 82:
```suggestion
    def url_parts = url.tokenize('/@')
    def user_info = url_parts[1].split(":")
    def user = user_info[0]
    def pass = user_info.length > 1 ? user_info[1] : null
    def (host, port) = url_parts[2].split(":")
    def database = url_parts.length > 3 ? url_parts[3] : null
```
This is just an alternative code to avoid too many splits and index manipulation, but happy to leave yours (although in that case I would suggest to change `urlparts` to `url_parts` and add spaces in `user_info[1] : null`).

JAlvarezJarreta (changes requested):
Couple of comments/suggestions, but otherwise looks good, thank you! :+1:

pipelines/nextflow/modules/utils/utils.nf
```diff
@@ -50,3 +50,38 @@ def parse_list_param(String multi_value = '', List<String> allowed_values, Boole
...
+    if (database) {
+        base_url += "/${database}"
+    }
+    return base_url
+}
+
+
+def extractMySQLArguments(urlString) {
+    //Extract the MySQL arguments from the url
+    def remove_protocol = urlString.split("//")[1]
```
MatBarba, line 70: Can you add a check for the protocol (make sure it is indeed "mysql")? Also a check that the URL format is correct. If any of those are missing, it should throw an exception.

pipelines/nextflow/modules/utils/utils.nf
```diff
@@ -50,3 +50,38 @@ def parse_list_param(String multi_value = '', List<String> allowed_values, Boole
...
+
+    def urlparts = remove_protocol.split("@")
+    def user_info = urlparts[0].split(":")
+    def user = user_info[0]
+    def pass = user_info.length > 1 ? user_info[1]:null
```
MatBarba, line 75:
```suggestion
    def pass = user_info.length > 1 ? user_info[1] : null
```

MatBarba (changes requested):

pipelines/nextflow/modules/utils/utils.nf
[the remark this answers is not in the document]
JAlvarezJarreta: Now Disha will need to decide between accepting your suggestion and my suggestion :smiling_imp:

JAlvarezJarreta (approved):
Nice update, cheers!

pipelines/nextflow/modules/utils/utils.nf
```diff
@@ -50,3 +50,33 @@ def parse_list_param(String multi_value = '', List<String> allowed_values, Boole
...
     return all_params
 }
+
+def generate_url(protocol, host, port, user, pass=null, database=null) {
+    // Generate a URL when all the credentials are provided
+    base_url = "${protocol}://${user}@${host}:${port}"
+
+    if (pass) {
+        base_url = "${protocol}://${user}:${pass}@${host}:${port}"
+    }
```
MatBarba, line 60: Better to use full name for consistency:
```suggestion
def generate_url(protocol, host, port, user, password=null, database=null) {
    // Generate a URL when all the credentials are provided
    base_url = "${protocol}://${user}@${host}:${port}"

    if (password) {
        base_url = "${protocol}://${user}:${password}@${host}:${port}"
    }
```

pipelines/nextflow/modules/utils/utils.nf
```diff
@@ -50,3 +50,33 @@ def parse_list_param(String multi_value = '', List<String> allowed_values, Boole
...
+      return [error: "Invalid url ${url_string}. A simple Mysql URL is expected."]
+    }
+    def url_parts = url_string.tokenize('/@')
+    def user_info = url_parts[1].split(":")
+    def user = user_info[0]
+    def pass = user_info.length > 1 ? user_info[1] : null
+    def (host, port) = url_parts[2].split(":")
+    def database = url_parts.size() > 3 ? url_parts[3] : null
+
+    return [user: user, pass: pass, host: host, port: port, database: database]
```
MatBarba, line 81: Processes expect `password`, also better use full name for consistency:
```suggestion
    def password = user_info.length > 1 ? user_info[1] : null
    def (host, port) = url_parts[2].split(":")
    def database = url_parts.size() > 3 ? url_parts[3] : null

    return [user: user, password: password, host: host, port: port, database: database]
```

MatBarba (approved):
Last small hitch then we should be good

PR #397, opened 2024-07-10
Remove unused dependency and only depend on PyPI libraries

JAlvarezJarreta:
This PR is needed because PyPI does not allow publication of libraries that depend on "indirect code", i.e. libraries not published in PyPI.


pyproject.toml
```diff
@@ -51,8 +51,7 @@ classifiers = [
 dependencies = [
     "bcbio-gff == 0.7.1",
     "biopython >= 1.81",
-    "ensembl-hive @ git+https://github.com/Ensembl/ensembl-hive.git",
```
ens-LCampbell, line 54: So we'd be leveraging ensembl-py dependency on ensembl-hive and removing this as a core dep on ensembl-genomio ? Whenever ense-py is in PyPI that is.

JAlvarezJarreta: No, basically this dependency is removed from the Python code, as this library is not required to be installed in a Python environment for users to be able to install and use ensembl-genomio. They need ensembl-hive to run our pipelines, but that's a different type of dependency (not to be added in the `pyproject.toml`).

ens-LCampbell: Ah I see, ok well that makes perfect sense thanks for clarifying !

ens-LCampbell (approved):

PR #399, opened 2024-07-15
Drop some brc4 modules

MatBarba:
Drop the following Hive pipelines, since they have been migrated to Nextflow + Python (or container scripts):
- addition prepare
- genome prepare
- dumper (need the container PR to be merged for that one)

All Python runnables used by those and not used by anything else are also removed.

I've also removed some Perl runnables in the same fashion.

We need to decide if we want to keep the genome compare pipeline, and the runnables would need to be migrated.

The only complicated pipeline remaining will be the genome loader pipeline (and associated LoadGFF3 pipelines).


JAlvarezJarreta (approved):
Nice first cleanup of almost 5k lines of code.

PR #400, opened 2024-07-16
Move singularity enabled in slurm/lsf profiles

MatBarba:
Only enable singularity on our cluster, i.e. when we use the profiles `slurm` or `lsf`, instead of systematically. This means that singularity is no longer enabled by default when run locally (by default without a `--profile`), in particular where it is not available (e.g. on a Mac).

NB: on Mac you can use Docker for processes that have a container, so you should add this to your `~/.nextflow/config`:
```
docker.enabled = true
```
(Unless you want to use you current environment)

This should make all our pipelines (and their tests) run on both clusters and laptops without changing the default configs

### DIFF
--- a/pipelines/nextflow/workflows/additional_seq_prepare/nextflow.config
+++ b/pipelines/nextflow/workflows/additional_seq_prepare/nextflow.config
@@ -24,7 +24,3 @@ params {
     prefix = null
     production_name = null
 }
-
-singularity {
-	enabled = true
-}
--- a/pipelines/nextflow/workflows/genome_prepare/nextflow.config
+++ b/pipelines/nextflow/workflows/genome_prepare/nextflow.config
@@ -23,7 +23,3 @@ params {
     output_dir = "output_genome_prepare"
     ncbi_check = 1
 }
-
-singularity {
-	enabled = true
-}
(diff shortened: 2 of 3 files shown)


JAlvarezJarreta (approved):

JAlvarezJarreta:
What about adding `autoMounts = true` for the singularity configuration?

ens-LCampbell:
> What about adding `autoMounts = true` for the singularity configuration? Or, alternatively, setting the minimum version to 23.09 instead of 23.01 [[link](https://www.nextflow.io/docs/latest/container.html#id23)].

We might as well increase the nextflow version to 23.09... however I don't see this on codon 23.04 is the max no ?

JAlvarezJarreta:
> We might as well increase the nextflow version to 23.09... however I don't see this on codon 23.04 is the max no ?

Yes, we can request that version to be installed (of 24.04?), but it may be quicker (and easier) to just add that configuration parameter.

ens-LCampbell:
> > We might as well increase the nextflow version to 23.09... however I don't see this on codon 23.04 is the max no ?
> 
> Yes, we can request that version to be installed (of 24.04?), but it may be quicker (and easier) to just add that configuration parameter.

I think add @MatBarba config changes, then we can update to later version. Might be better to add 23.09 on top of 23.04 or over one of the oldest installed versions

ens-LCampbell (approved):

PR #401, opened 2024-07-16
Move gt container in main config

MatBarba:
Centralize the containers url in the nextflow.config file. This makes it easier to reuse the containers in various processes using labels if needed, and it's easier to maintain in the long run.

### DIFF
--- a/pipelines/nextflow/modules/gff3/gff3_validation.nf
+++ b/pipelines/nextflow/modules/gff3/gff3_validation.nf
@@ -15,10 +15,9 @@
 
 process GFF3_VALIDATION {
 
-  //beforeScript 'module load libffi-3.3-gcc-9.3.0-cgokng6'
   tag "${gene_models}"
   label 'adaptive'
-  container "quay.io/biocontainers/genometools-genometools:1.6.5--py310h3db02ab_0"
+  label 'container_genometools'
 
   input:
     tuple val(meta), path (gene_models, stageAs: "incoming.gff3")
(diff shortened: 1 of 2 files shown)


JAlvarezJarreta (approved):

ens-LCampbell (approved):
Totally, this is how I have setup Datasets container (upcoming PR). 

Just one thing, do we want to perhaps host this container within dockerhub ensemblorg instead of quay ? 

Under umbrella a new genomio project ? Im sure we would have some additional misc containers/packages we might need in the future which would also go here. Just a thought

MatBarba:
> Totally, this is how I have setup Datasets container (upcoming PR).
> 
> Just one thing, do we want to perhaps host this container within dockerhub ensemblorg instead of quay ?
> 
> Under umbrella a new genomio project ? Im sure we would have some additional misc containers/packages we might need in the future which would also go here. Just a thought

Possibly yes. We can also use the container I made for the dumpers as it also contains genometools, but the same question of hosting by Ensembl arises.

JAlvarezJarreta:
> Possibly yes. We can also use the container I made for the dumpers as it also contains genometools, but the same question of hosting by Ensembl arises.

Whilst I may make sense to have such a combined container to simplify the pipeline and make a process atomic, due to the added maintenance cost and complexity of combining several tools in containers (among others, us having to maintain a container that may be broken due to an issue with `genometools`, which is out of our control), it may be better in the long term to split the process in two, so it only needs one container or the other.

PR #402, opened 2024-07-17
Update nf-validation to nf-schema

MatBarba:
Use fixed version to prevent plugin check every time we run the pipeline (in particular when running tests)


JAlvarezJarreta (approved):
I'm surprised we are 100% compatible with v2.0, they claimed it was not backwards compatible.

MatBarba:
> I'm surprised we are 100% compatible with v2.0, they claimed it was not backwards compatible.

I guess we're lucky we did not use those things that broke (I think we just did not use anything advanced enough for that)

PR #403, opened 2024-07-17
Set same behaviour as later Nextflow versions (23.05+)

JAlvarezJarreta:
I'm also considering dropping support for LSF since we should be using only SLURM now, but will leave that decision to the reviewer.

### DIFF
--- a/pipelines/nextflow/workflows/nextflow.config
+++ b/pipelines/nextflow/workflows/nextflow.config
@@ -57,6 +57,7 @@ profiles {
         }
         singularity {
             enabled = true
+            autoMounts = true
         }
     }
 
@@ -72,6 +73,7 @@ profiles {
         }
         singularity {
             enabled = true
+            autoMounts = true
         }
     }
 


MatBarba (commented):
What does this change do?

MatBarba:
About LSF: let's keep it until LSF is actually retired

JAlvarezJarreta:
> What does this change do?

Nextflow does not automatically mount the required host paths in the container. This behaviour is changed from v23.09, so this way we have the same behaviour now as we will have afterwards (so we can detect and fix any bugs before we move to a newer version).

MatBarba:
> > What does this change do?
> 
> Nextflow does not automatically mount the required host paths in the container. This behaviour is changed from v23.09, so this way we have the same behaviour now as we will have afterwards (so we can detect and fix any bugs before we move to a newer version).

What does that mean exactly? That we should be able to access any data (and program) on the host system? I'm trying to understand in what cases we would do that.

MatBarba (approved):
Should not break anything in principle since we don't use that feature right now

JAlvarezJarreta:
> What does that mean exactly? That we should be able to access any data (and program) on the host system? I'm trying to understand in what cases we would do that.

I believe so, although I'm not 100% sure either. It is more to align with coming behaviour than anything else.

PR #408, opened 2024-07-19
Convert chunk fasta script to module and remove gff3_extract_annotation entry point

ens-LCampbell:
1. Added a new chunk sub module to the fasta module
Source of said submodule was original stand alone script in scripts folder
Added some small tweaks, logging, and integration of ensembl-utils for file extraction. 

2. Removed broken entry point to non-standalone submodule in gff3: extract_annotation.
This entry point was directing to module without a main function. 

3. Updated scripts/Readme

4. Specified UTF-8 encoding in utils/json_utils


src/python/ensembl/io/genomio/fasta/chunk.py
```diff
@@ -14,48 +14,99 @@
...
     )
-    # out
     parser.add_argument(
-        "--out", metavar="chunks.fna", required=False, default=None, type=str, help="chunks output [STDOUT]"
+        "--out",
+        metavar="chunks.fna",
+        required=False,
+        default="chunks.fna",
+        type=str,
+        help="chunks output [STDOUT]",
```
JAlvarezJarreta, line 96: Cannot change this, but I would use `parser.add_argument_dst_path(` instead of `parser.add_argument(`.
```suggestion
        "--out",
        required=False,
        default="chunks.fna",
        help="Chunks output file",
```

src/python/ensembl/io/genomio/fasta/chunk.py
```diff
@@ -14,48 +14,99 @@
...
+            # exclude starts, as they are 0 or pushed as previous chunk_ends
+            result += list(range(offset, chunk_end, chunk_size))[1:]
+        result.append(chunk_end)
+        offset = chunk_end
+    return result
+
+
+## MAIN ##
+def main():
+    """Modules entry-point."""
```
JAlvarezJarreta, line 82:
```suggestion
    """Module's entry-point."""
```

src/python/ensembl/io/genomio/fasta/chunk.py
```diff
@@ -64,78 +115,48 @@ def get_args():
...
+            it will not be split.",
     )
     parser.add_argument(
         "--n_seq",
         required=False,
         default=0,
         type=int,
-        help="The fasta file will be split into chunks at positions of at least n N sequences. n number is specified here.",
+        help="The fasta file will be split into chunks at positions of at \
+            least n N sequences. n number is specified here.",
```
JAlvarezJarreta, line 142:
```suggestion
        help="Split into chunks at positions of at least this number of N characters.",
```

src/python/ensembl/io/genomio/fasta/chunk.py
```diff
@@ -64,78 +115,48 @@ def get_args():
...
     )
-    parser.add_argument("--add_offset", required=False, action="store_true", help="0 based offset")
+    parser.add_argument("--add_offset", required=False, action="store_true", help="0 based offset.")
```
JAlvarezJarreta, line 144:
```suggestion
    parser.add_argument("--add_offset", required=False, action="store_true", help="Zero-based offset.")
```

src/python/ensembl/io/genomio/fasta/chunk.py
```diff
@@ -153,25 +174,22 @@ def main():
...
         )
 
         fasta_parser = SeqIO.parse(fasta, "fasta")
         for rec_count, rec in enumerate(fasta_parser, start=1):
             rec_name = str(rec.name)
-            rec_len = len(rec)
 
-            ends = split_by_n(str(rec.seq), n_split_regex)  # returns [ len(req.seq) ] if no regexp
-            ends = split_by_chunk_size(ends, args.chunk_size, tolerated_chunk_len)
+            ends = split_seq_by_n(str(rec.seq), n_split_regex)  # returns [ len(req.seq) ] if no regexp
```
JAlvarezJarreta, line 191:
```suggestion
            ends = split_seq_by_n(str(rec.seq), n_split_regex)
```

JAlvarezJarreta (changes requested):
Couple of general requests before digging more into the code:
- Since the code in `main()` functions cannot be tested, can you please move it to a separate function?
- Can you replace the use of `os.path` by `pathlib.Path`?

MatBarba:
> Couple of general requests before digging more into the code:
> 
>     * Since the code in `main()` functions cannot be tested, can you please move it to a separate function?
> 
>     * Can you replace the use of `os.path` by `pathlib.Path`?

I would recommend we do that in a separate PR. We should separate the code conversion (already pretty complicated/confusing) and code improvement (which could result in a code overhowl, so a lot of separate changes on top of the conversion).

I.e. we know the code "works" in the script, so we should first convert it to a genomio module with minimal changes. Then only in a second step should we make actual coding changes.

MatBarba (approved):
Looks good! We can make some improvements to better follow guidelines, and add tests etc. but this can be done separately IMO

src/python/ensembl/io/genomio/fasta/chunk.py
ens-LCampbell: I believe this suggestion isn't correct. The plural of module modules. Module's == module is for short. I just will remove the s entirely

src/python/ensembl/io/genomio/fasta/chunk.py
```diff
@@ -14,128 +14,146 @@
...
     parser.add_argument(
         "--individual_out_dir",
         required=False,
         default=None,
-        type=str,
-        help="output directory for writing files with individual chunks to (optional), "
-        "if provided,`--out` value used as a filename prefix",
+        type=Path,
+        help="Output directory for writing files with individual chunks to. \
+        If provided,`--out` value used as a filename prefix",
```
JAlvarezJarreta, line 102:
```suggestion
            If provided,`--out` value used as a filename prefix",
```

JAlvarezJarreta (changes requested):
Looks good to me.

JAlvarezJarreta (approved):

PR #413, opened 2024-07-31
Mbarba/simpler acc

MatBarba:
- Only keep one version of `download_genome_meta` by getting the accession from the `db` earlier (already available in the db meta).
- Use a meta value for all processes, with expected keys id (for display/uniqueness) and accession (INSDC/RefSeq)
- Generation of the meta values is clearer


pipelines/nextflow/modules/schema/check_json_schema_db.nf
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import java.io.File
+// import java.io.File
```
ens-LCampbell, line 16: Do we still need this line ?

ens-LCampbell (approved):
Overall I think a much cleaner implementation, thanks for unifying those datasets modules nice work ! 👍 

An aside: Do we need to integrate these changes into genome_prepare pipeline ?

pipelines/nextflow/modules/schema/check_json_schema_db.nf
MatBarba: no indeed

JAlvarezJarreta (approved):
Couple of tiny details, but looks good otherwise. Like the improvements, definitely something worth pursuing in every pipeline/module.

PR #415, opened 2024-08-06
Fix process name and mysql arguments

MatBarba:
Proper name for the process and use correct mysql arguments, for the patch build pipeline.

### DIFF
--- a/pipelines/nextflow/modules/patch_build/canonical_transcripts.nf
+++ b/pipelines/nextflow/modules/patch_build/canonical_transcripts.nf
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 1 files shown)


pipelines/nextflow/modules/patch_build/canonical_transcripts.nf
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-process CHECK_PATCH {
+process CANONICAL_TRANSCRIPTS {
```
JAlvarezJarreta, line 16: Not too relevant in this case, but processes should start with a verb (not asking to change it in this case).

JAlvarezJarreta (approved):

PR #421, opened 2024-08-14
Drop seq_region_name script, obsolete

MatBarba:
This script was made to help update the name of seq_region BRC names, but it's no longer in use (not used in pipelines either).


JAlvarezJarreta (approved):
Always good to remove deprecated things :+1: 

`ensembl-genomio/src/python/ensembl/io/genomio/seq_region/__init__.py` needs to be updated accordingly though.

PR #426, opened 2024-08-19
Fix assert_called

MatBarba:
After having rebuild my venv, I encountered an issue with these `assert_call` calls. I've made a quick fix but we should look at this test more closely to make sure it's correctly testing with these changes.


ens-LCampbell (commented):
So this change is a stop gap solution to allow tests to pass is that right ?

MatBarba:
> So this change is a stop gap solution to allow tests to pass is that right ?

That's right

PR #427, opened 2024-08-23
Upgrade to ensembl-py 2.1

MatBarba:
Make various fixes to ensure we can upgrade to ensembl-py 2.1, which allows using SQLalchemy 2.0 and not just 1.4

Main changes are to the tests for genome_stats/dump and core_server:
- `func.count` is flagged erronously by pylint as `not-callable` (cf https://github.com/sqlalchemy/sqlalchemy/issues/9189), so added an ignore comment
- Use `func.count()` with argument (not needed)
- Replace session->execute() with session->connection->execute() (with an additional MockConnection)
- Mypy does not like how we overload `MockSession` with a different `execute()`, so added mypy ignore override

PS: We should plan to rewrite the genome_stats/dump with UniTestDB instead to avoid those mypy/pylint flags

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ classifiers = [
 dependencies = [
     "bcbio-gff == 0.7.1",
     "biopython >= 1.81",
-    "ensembl-py >= 2.1",
+    "ensembl-py >= 2.1.0",
     "ensembl-utils >= 0.4.1",
     "jsonschema >= 4.6.0",
     "intervaltree >= 3.1.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ classifiers = [
 dependencies = [
     "bcbio-gff == 0.7.1",
     "biopython >= 1.81",
-    "ensembl-py >= v2.0.2",
+    "ensembl-py >= 2.1",
     "ensembl-utils >= 0.4.1",
     "jsonschema >= 4.6.0",
     "intervaltree >= 3.1.0",
--- a/src/python/ensembl/io/genomio/database/core_server.py
+++ b/src/python/ensembl/io/genomio/database/core_server.py
@@ -22,6 +22,7 @@ import logging
 
 import sqlalchemy
 from sqlalchemy.engine import URL
+from sqlalchemy import text
 
 
 class CoreServer:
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -51,7 +51,7 @@ classifiers = [
 dependencies = [
     "bcbio-gff == 0.7.1",
     "biopython >= 1.81",
-    "ensembl-py >= v2.0.2",
+    "ensembl-py >= 2.1",
```
JAlvarezJarreta, line 54:
```suggestion
    "ensembl-py >= 2.1.0",
```

src/python/ensembl/io/genomio/database/core_server.py
```diff
@@ -37,8 +38,9 @@ def __init__(self, server_url: URL) -> None:
     def get_all_core_names(self) -> List[str]:
         """Query the server and retrieve all database names that look like Ensembl cores."""
 
-        all_query = self.engine.execute(r"SHOW DATABASES LIKE '%%_core_%%'")
-        dbs = [row[0] for row in all_query.fetchall()]
+        with self.engine.connect() as connection:
+            all_query = connection.execute(text(r"SHOW DATABASES LIKE '%%_core_%%'"))
```
JAlvarezJarreta, line 42: Not sure if `r` before the string is necessary when using `text()`.

src/python/tests/database/test_core_server.py
```diff
@@ -47,19 +48,36 @@ def __init__(self, core_dbs: List[str]):
         self.core_dbs = core_dbs
 
     def fetchall(self) -> List[List[str]]:
-        """Return a list of lists, ech one containing a single core db."""
+        """Return a list of lists, each one containing a single core db."""
         return [[x] for x in self.core_dbs]
 
 
+class MockConnection:
+    """Mock a SQLalchemy connection."""
```
JAlvarezJarreta, line 56:
```suggestion
    """Mock a SQLAlchemy connection."""
```

src/python/tests/database/test_core_server.py
```diff
@@ -47,19 +48,36 @@ def __init__(self, core_dbs: List[str]):
...
+
+    def __init__(self, result: MockResult) -> None:
+        self.result = result
+
+    def execute(self, *args, **kwargs) -> MockResult:  # pylint: disable=unused-argument
+        """Returns a MockResult object."""
```
JAlvarezJarreta, line 62:
```suggestion
        """Returns a `MockResult` object."""
```

JAlvarezJarreta (approved):
Everything looks sensible to me, besides the significant addition of ignore bits, but I assume SQLAlchemy is still improving so those will be needed for a while.

A couple of minor suggestions, that's all.

JAlvarezJarreta (approved):

PR #431, opened 2024-09-05
Require type hints in every method and function

JAlvarezJarreta:
On top of the added configuration of `mypy`:
- Updated codebase accordingly
- Fixed some typos
- Updated some docstrings

It is interesting how forcing this has not only flagged a few things we missed during code review, but also a few minor issues we had in the code (and most likely a couple of bugs).


vsitnik (approved):
cool. Thanks a lot! Should be ok.

PR #437, opened 2024-09-12
Upgrade ensembl-py to 2.1.2 to fix seq_region dump bug

MatBarba:
Fix ensembl-py ORM to allow dumps from MySQL core dbs.
Also upgrade ensembl-utils to its latest version.


ens-LCampbell (approved):
Approved thanks dude !!

PR #438, opened 2024-09-12
Revert relative paths to absolute paths so they work properly in pypi

JAlvarezJarreta:
It seems there is no alternative solution to this problem, and it is better to have working links in `pypi` than links working for every branch in the repository.


ens-LCampbell (approved):
Annoying, seems like a feature that could be accommodated. At least the URLs won't change on main side.... until they a year or two down the road ha

PR #443, opened 2024-09-18
Old brc4 minor fixes

MatBarba:


JAlvarezJarreta (approved):

src/perl/Bio/EnsEMBL/Pipeline/PipeConfig/BRC4_genome_compare_conf.pm
```diff
@@ -156,7 +156,6 @@ sub pipeline_analyses {
     { -logic_name  => 'Fasta_DNA',
       -module      => 'Bio::EnsEMBL::Pipeline::Runnable::BRC4::DumpFastaDNA',
       -parameters => {
-        hash_key => "insdc_fasta_dna",
```
Dishalodha, line 159: Good spot, however I think we need to make this change in the first tag we created as the pipeline is already not functional in the current branch because some python scripts have been moved.

Dishalodha (approved):
Happy to merge this! We need to update or retire the compare genomes pipeline later

src/perl/Bio/EnsEMBL/Pipeline/PipeConfig/BRC4_genome_compare_conf.pm
MatBarba: Ah maybe this PR should be moved to be against the legacy branch instead?

Dishalodha: Maybe we will merge it to the main branch for now and the legacy branch can be changed in a separate PR

PR #444, opened 2024-09-24
Remove patch build nextflow modules

JAlvarezJarreta:
With VEuPathDB concluded and its code being tagged (and saved in a specific branch), we can start dropping those elements that are no longer needed in our codebase. This is the first step.

Note that the remove of this module has also forced to drop the patch-build Nextflow pipeline, as most elements were being removed in the process.


Dishalodha (approved):
Happy to remove these, thank you

PR #447, opened 2024-10-04
Use parents feature instead of loop for subparsers

JAlvarezJarreta:
I was the one proposing the addition of common arguments via loop for each subparser, but reviewing the documentation I have found the "correct" way to implement it, so I have fixed it in the two places where we had it.

### DIFF
--- a/src/python/ensembl/io/genomio/assembly/status.py
+++ b/src/python/ensembl/io/genomio/assembly/status.py
@@ -346,9 +346,46 @@ def generate_report_tsv(
 def main() -> None:
     """Module's entry-point."""
     parser = ArgumentParser(description=__doc__)
+    # Create parser with common arguments to be used by both subparsers
+    base_parser = ArgumentParser(add_help=False)
+    base_parser.add_argument_dst_path(
+        "--output_file",
+        default=Path("assembly_report_jsons"),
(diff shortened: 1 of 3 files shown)


src/python/ensembl/io/genomio/assembly/status.py
```diff
@@ -346,55 +346,59 @@ def generate_report_tsv(
...
+        metavar="BATCH_SIZE",
+        help="number of accessions requested in one query to datasets",
+    )
+    base_parser.add_log_arguments(add_log_file=True)
+    # Add subparsers with their parent being the base parser with the common arguments
     subparsers = parser.add_subparsers(title="report assembly status from", required=True, dest="src")
     # Specific arguments required when using Ensembl core database names as source
-    core_db_parser = subparsers.add_parser("core_db", help="list of Ensembl core databases")
+    core_db_parser = subparsers.add_parser(
+        "core_db", parents=[base_parser], help="list of Ensembl core databases"
```
ens-LCampbell, line 387: Ahh ha nice trick

ens-LCampbell (approved):
So you've had to reorder the optional sub parsers to inherit from the main base parser, nice fix looks good to me

ens-LCampbell (approved):
if mypy and pylint are happy Im happy with merge :)

PR #448, opened 2024-10-07
Bugfix and output param name change

ens-LCampbell:
There was a bug with the use of the download dir path for datasets cli. 
- In place was param output_file, but it was not update in two function calls. 
- I've change the name of this param to better represent the actual function, its not a file naming param after all. Designating this param as 'datasets_dir' makes more sense to me.


JAlvarezJarreta (approved):

JAlvarezJarreta:
Good patch and nice param naming 😉

PR #449, opened 2024-10-07
Update database.factory.main() function to test all its contents

JAlvarezJarreta:
This is a showcase of how we could refactor our `main()` methods if we want to follow a path where logic can be left in `main()` but testing everything, argument parsing included (without having to go to the extent of using `subprocess`).

### DIFF
--- a/src/python/ensembl/io/genomio/database/factory.py
+++ b/src/python/ensembl/io/genomio/database/factory.py
@@ -20,7 +20,6 @@ import argparse
 import json
 import logging
 from pathlib import Path
-import sys
 
 from sqlalchemy.engine import URL
 
--- a/src/python/ensembl/io/genomio/database/factory.py
+++ b/src/python/ensembl/io/genomio/database/factory.py
@@ -16,10 +16,11 @@
 
 __all__ = ["format_db_data", "get_core_dbs_metadata"]
 
+import argparse
 import json
-from pathlib import Path
-from typing import Dict, List, Optional
 import logging
+from pathlib import Path
+import sys
 
 from sqlalchemy.engine import URL
 
@@ -29,7 +30,7 @@ from .core_server import CoreServer
 from .dbconnection_lite import DBConnectionLite
 
 
-def format_db_data(server_url: URL, dbs: List[str], brc_mode: bool = False) -> List[Dict]:
+def format_db_data(server_url: URL, dbs: list[str], brc_mode: bool = False) -> list[dict]:
     """Returns a metadata list from the given databases on a server.
 
     Args:
@@ -87,13 +88,14 @@ def format_db_data(server_url: URL, dbs: List[str], brc_mode: bool = False) -> L
 
 def get_core_dbs_metadata(
     server_url: URL,
+    *,
     prefix: str = "",
-    build: Optional[int] = None,
-    version: Optional[int] = None,
+    build: int | None = None,
+    version: int | None = None,
     db_regex: str = "",
-    db_list: Optional[Path] = None,
+    db_list: Path | None = None,
     brc_mode: bool = False,
-) -> List[Dict]:
+) -> list[dict]:
     """Returns all the metadata fetched for the selected core databases.
 
     Args:
@@ -123,8 +125,12 @@ def get_core_dbs_metadata(
     return format_db_data(server_url, databases, brc_mode)
 
 
-def main() -> None:
-    """Main script entry-point."""
+def parse_args(arg_list: list[str] | None) -> argparse.Namespace:
+    """TODO
+
+    Args:
+        arg_list: TODO
+    """
     parser = ArgumentParser(description=__doc__)
     parser.add_server_arguments()
     # Add filter arguments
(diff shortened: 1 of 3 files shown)


JAlvarezJarreta:
As a side effect, our code coverage would currently drop from 76% to 72% due to all the `main()` methods without unit test.

src/python/ensembl/io/genomio/database/factory.py
```diff
@@ -140,7 +145,17 @@ def main() -> None:
         help="Enable BRC mode, i.e. use organism_abbrev for species, component for division",
     )
     parser.add_log_arguments()
-    args = parser.parse_args()
+    return parser.parse_args(arg_list)
+
+
+def main(arg_list: list[str] | None = None) -> None:
```
vsitnik, line 151: why do we need arg_list?

Dishalodha (approved):

src/python/ensembl/io/genomio/database/factory.py
```diff
@@ -87,13 +87,14 @@ def format_db_data(server_url: URL, dbs: List[str], brc_mode: bool = False) -> L
 
 def get_core_dbs_metadata(
     server_url: URL,
+    *,
```
Dishalodha, line 90: for my understanding, why do we need to specify this?

JAlvarezJarreta: This forces all the arguments after the `*` to be keyworded, e.g. if you were to do `get_core_dbs_metadata("mysql://user@localhost:42", "my_prefix")`, the code will raise an error as the second parameter needs to have a keyword associated to it. `get_core_dbs_metadata("mysql://user@localhost:42", prefix="my_prefix")` would be the correct code.

src/python/ensembl/io/genomio/database/factory.py
JAlvarezJarreta: This was suggested by the page I was following to understand this approach. As far as I can tell, it mainly allows you for an easier unit testing of the `main()` method and reuse of the argument parsing declaration. Not my favourite implementation, but serves its purpose as PoC.

src/python/ensembl/io/genomio/database/factory.py
```diff
@@ -87,13 +87,14 @@ def format_db_data(server_url: URL, dbs: List[str], brc_mode: bool = False) -> L
...
     prefix: str = "",
-    build: Optional[int] = None,
-    version: Optional[int] = None,
+    build: int | None = None,
+    version: int | None = None,
```
ens-LCampbell, line 93: Are you opting to follow [PEP604](https://peps.python.org/pep-0604/) ?

src/python/tests/database/test_factory.py
```diff
@@ -169,7 +169,7 @@ def test_get_core_dbs_metadata(
     mock_format_db_data: Mock,
     data_dir: Path,
     use_db_file: bool,
-    output: List[Dict],
+    output: list[dict],
```
ens-LCampbell, line 172: I can't keep up with convention on this change from List[Dict] to list[dict] it seems to flip flop, or at least I struggle to recall which is preferred

src/python/tests/database/test_factory.py
```diff
@@ -196,3 +196,106 @@ def _format_db_data(server_url: URL, dbs: List[str], brc_mode: bool = False) ->
...
+    capsys: CaptureFixture[str],
+    arg_list: list[str],
+    server_url: URL,
+    stdout: str,
+) -> None:
+    """Tests the `factory.main()` function (entry point).
+
+    Fixtures: capsys
+    """
+    mock_get_core_dbs_metadata.return_value = {"test": "output"}
```
ens-LCampbell, line 293: I think Im following what this test does, but would like a little help understanding at next dev meeting

ens-LCampbell (commented):
All in all appreciate the effort to tackle the problem of main param passing testing. Happy for further discussion soon.

src/python/tests/database/test_factory.py
JAlvarezJarreta: So `List[Dict]` was the only option pre-Python 3.9, but then they changed the type hints to allow use of `list` and `dict` (among other things). I'm just applying it when I see it may help me remove the import line (one line less of code, yay!). Not important, should not be done overall unless you find yourself with time. In this case, since I was showcasing a PoC, felt it was worth showing this as well.

src/python/ensembl/io/genomio/database/factory.py
JAlvarezJarreta: Yes, this is a change on type hints that came with Python 3.9 if I recall correctly. I find those updates a little bit more readable and sensible than before. _See next comment for more context and opinions on this_

src/python/tests/database/test_factory.py
```diff
@@ -20,13 +20,13 @@
...
 
 from pathlib import Path
-from typing import Dict, List, Optional
 from unittest.mock import call, Mock, patch
 
 from deepdiff import DeepDiff
 import pytest
 from pytest import param
-from sqlalchemy.engine import URL
+from _pytest.capture import CaptureFixture
```
ens-LCampbell, line 28: Could you explain at dev meeting why you have both pytest and _pytest imports ?

ens-LCampbell (approved):

ens-LCampbell (approved):

PR #450, opened 2024-10-24
Update GenomIO to avoid too many positional arguments

JAlvarezJarreta:
But ignore the issue in unit tests, since it is not relevant in that case.

### DIFF
--- a/src/python/tests/fasta/test_chunk.py
+++ b/src/python/tests/fasta/test_chunk.py
@@ -316,9 +316,9 @@ def test_chunk_fasta_stream(
                     chunk_size_tolerated,
                     output_fasta,
                     None,
-                    n_sequence_len,
-                    chunk_sfx,
-                    append_offset_to_chunk_name,
+                    n_sequence_len=n_sequence_len,
+                    chunk_sfx=chunk_sfx,
+                    append_offset_to_chunk_name=append_offset_to_chunk_name,
                     open_individual=_individual_opener,
                 )
             assert output_fasta.getvalue() == expected_chunked_fasta_text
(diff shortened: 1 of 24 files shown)


src/python/tests/fasta/test_chunk.py
```diff
@@ -12,12 +12,8 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Unit testing of `ensembl.io.genomio.fasta.chunk` module.
-
-Typical usage example::
-    $ pytest test_chunk.py
-
-"""
+"""Unit testing of `ensembl.io.genomio.fasta.chunk` module."""
+# pylint: disable=too-many-positional-arguments
```
ens-LCampbell, line 16: Slightly confused on adding this comment line, does this disable TMPA ?

ens-LCampbell: Obvs same applies to other insertions of this line, just to clarify thanks

ens-LCampbell (approved):
Looks solid thanks Jorge :) Approved, just wanting clarification on tests pylint comment lines added

src/python/tests/fasta/test_chunk.py
JAlvarezJarreta: Yes, this disables that particular check for the whole file. I would have preferred to disable it completely for the `tests` folder, but `pylint` does not have such feature, so it has to be manually introduced in each test file.

ens-LCampbell: Oh ok ok just checking as I wasn't sure if that enabled or disabelled the check thanks for clearing that up. Merge approved from my end

PR #455, opened 2024-11-19
Adding CICD trigger to allow non versioned release genomio container builds 

ens-LCampbell:
This PR is simply attempting to allow genomio maintainers to decouple the creation of a genomio python container from the current paradigm of A (Create versioned release) -> trigger CICD docker generation and publish to gitlab + dockerhub. 

If a user wants to for whatever reason rebuild the docker container (in this case we needed to rebuild as the container was missing 'ps') then commits can be made, then tagged with a format specific _[git tag](https://docs.gitlab.com/ee/user/project/repository/tags/)_. This should (in principle), though as of now is untested, allow for manual triggering of the pipeline via selection of said git tag. 

**Tag format:** 
GenomioDockerRebuild_v[0-9]+\.[0-9]+\.[0-9]+[a-z]$

In this case, the first ever manual trigger is named: GenomioDockerRebuild_v1.5.0a

I wanted to keep the genomio version information intact as this should aid future tag creation and history tracing. 

Users wanting to create a new tag for the above purpose of CICD pipeline initiation should follow the format and check before creation using `git tag -l --merged`. If a second tag is needed on the same version of genomio (i.e. release version), then serially increment the suffix **letter** a-> b-> c-> d... etc.

### DIFF
--- a/cicd/gitlab/dot.gitlab-ci.yml
+++ b/cicd/gitlab/dot.gitlab-ci.yml
@@ -54,6 +54,8 @@ include:
     rules:
       - if: '$CI_COMMIT_TAG =~ /v[0-9]+\.[0-9]+\.[0-9]+$/'
         when: always
+      - if: '$CI_COMMIT_TAG =~ /container_rebuild_[0-9]+$/'
+        when: always
 
 
 # External pipelines
--- a/cicd/gitlab/dot.gitlab-ci.yml
+++ b/cicd/gitlab/dot.gitlab-ci.yml
@@ -54,7 +54,7 @@ include:
     rules:
       - if: '$CI_COMMIT_TAG =~ /v[0-9]+\.[0-9]+\.[0-9]+$/'
         when: always
-      - if: '$CI_COMMIT_TAG =~ /container_rebuild_[0-9]+$/'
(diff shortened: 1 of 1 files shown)


Dishalodha (approved):
Happy with this to update the tag

cicd/gitlab/dot.gitlab-ci.yml
```diff
@@ -54,6 +54,8 @@ include:
     rules:
       - if: '$CI_COMMIT_TAG =~ /v[0-9]+\.[0-9]+\.[0-9]+$/'
         when: always
+      - if: '$CI_COMMIT_TAG =~ /GenomioDockerRebuild_v[0-9]+\.[0-9]+\.[0-9]+[a-z]$/'
```
vsitnik, line 57: do you need versions in the regexp at all?

vsitnik (approved):
thank you!

cicd/gitlab/dot.gitlab-ci.yml
ens-LCampbell: Yes Imho we need it, as it will be added to the docker tag as well as for tracking manual tags made with git tag. Also its needed to understand which genomio python version the container  actually is.

ens-LCampbell:
Thank you guys, will update on the success or lack thereof of this approach.

PR #456, opened 2024-12-11
Add --version to every entry point

JAlvarezJarreta:
It turns out Python's argument parser has a quick workaround for this, so it was pretty quick to update all our entry points.


vsitnik:
why not a single entry point instead?

JAlvarezJarreta:
> why not a single entry point instead?

I felt this was more the standard approach for every tool we use, even for tool suites such as `bcftools`. When checking how this should be done for Python libraries, the official answers are either use the `python -c "import ..."` approach we already have, or get it via `pip`, and both have their issues, as we know.

It is also only adding two lines of code, and we could even reduce it to one if we move the logic into `ensembl.utils`. Happy with what the majority decides, as I have no strong feelings for any solution.

ens-LCampbell (approved):
Looks good to me, thanks Jorge. Tested some entry points too and looks fine !

ens-LCampbell:
Side note, is there currently way quick way to list or find all genomio entry points available ? Would it be useful to add such a functionality and a single entry point cmd to list all (+ maybe add current library version as well) ?

JAlvarezJarreta:
> Side note, is there currently way quick way to list or find all genomio entry points available ? Would it be useful to add such a functionality and a single entry point cmd to list all (+ maybe add current library version as well) ?

The only way I can think of is looking at `pyproject.toml`. After some digging, there is a programmatic way to obtain them in Python:
```python
from importlib.metadata import entry_points
genomio_scripts = [script.name for script in entry_points(group='console_scripts') if script.module.startswith("ensembl.io.genomio.")]
print("\n".join(genomio_scripts))
```

Dishalodha (approved):
Looks good, a nice way of using version with the parser

ens-LCampbell:
@JAlvarezJarreta Should we also bump up the genomio version in this PR ?

vsitnik (approved):
Cool. Thanks!

PR #457, opened 2024-12-11
Update commit tag format when rebuilding docker containers

JAlvarezJarreta:
The new format will be, for instance, `v1.5.0_rebuild_a`, which is shorter and starts with the version for better tracking.

### DIFF
--- a/cicd/gitlab/dot.gitlab-ci.yml
+++ b/cicd/gitlab/dot.gitlab-ci.yml
@@ -52,9 +52,7 @@ include:
   # Genomio docker deploy (on release tag creation)
   - local: cicd/gitlab/parts/dockerbuild.genomio.gitlab-ci.yml
     rules:
-      - if: '$CI_COMMIT_TAG =~ /v[0-9]+\.[0-9]+\.[0-9]+$/'
-        when: always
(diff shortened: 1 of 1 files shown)


cicd/gitlab/dot.gitlab-ci.yml
```diff
@@ -52,9 +52,7 @@ include:
   # Genomio docker deploy (on release tag creation)
   - local: cicd/gitlab/parts/dockerbuild.genomio.gitlab-ci.yml
     rules:
-      - if: '$CI_COMMIT_TAG =~ /v[0-9]+\.[0-9]+\.[0-9]+$/'
```
ens-LCampbell, line 55: Im pretty sure we need to base tag as we need this to run when we create a standard release via merge to main + create a new release tag

ens-LCampbell: On closer inspection, the rebuild portion of this tag can be present or absent right. If so disregard first comment :)

JAlvarezJarreta: The new regex should cover both cases (the `_rebuild_[az]` is surrounded by a `?`). If it is not too clear, happy to divide it in two as it was before.

ens-LCampbell (approved):

PR #463, opened 2025-01-21
week long resource classes and rc_class params for BRC4_genome_loader

vsitnik:


JAlvarezJarreta (approved):
Nice addition, thank you!

PR #468, opened 2025-02-21
Remove old entry point left by mistake

JAlvarezJarreta:


ens-LCampbell (approved):
Haunting leftovers from days of yore 👍

PR #476, opened 2025-09-01
Disha/compare fasta

Dishalodha:
Updated compare fasta from runnables into a module. This module is mainly focused on comparing 2 fasta files, mainly with INSDC

- Updated the logic for comparing Ns in fasta file to identify the low quality data
- Removed the comparison for seq_region.json file

its a split from the PR #461

### DIFF
--- a/src/python/ensembl/io/genomio/fasta/compare.py
+++ b/src/python/ensembl/io/genomio/fasta/compare.py
@@ -162,8 +162,8 @@ class CompareFasta:
         """Find common sequences between two dictionaries and group them.
 
         Args:
-            seq_ext_dict (dict): Dictionary of sequences from the first dataset.
-            seq_core_dict (dict): Dictionary of sequences from the second dataset.
+            seq_ext_dict: Dictionary of sequences from the first dataset.
+            seq_core_dict: Dictionary of sequences from the second dataset.
 
         Returns:
             A dictionary of common sequence mappings and a list of comparison results.
--- a/src/python/ensembl/io/genomio/fasta/compare.py
+++ b/src/python/ensembl/io/genomio/fasta/compare.py
@@ -31,12 +31,7 @@ from ensembl.utils.logging import init_logging_with_args
 
 __all__ = [
     "SeqGroup",
-    "compare_seqs",
-    "read_fasta",
-    "build_seq_dict",
-    "find_common_groups",
-    "write_results",
-    "compare_seq_for_Ns",
+    "CompareFasta",
 ]
 
 
--- a/src/python/ensembl/io/genomio/fasta/compare.py
+++ b/src/python/ensembl/io/genomio/fasta/compare.py
@@ -29,7 +29,15 @@ from ensembl.utils.logging import init_logging_with_args
 
 """Compare sequences between two genomes"""
 
-__all__ = ["SeqGroup", "compare_seqs", "read_fasta", "build_seq_dict", "find_common_groups", "write_results", "compare_seq_for_Ns"]
+__all__ = [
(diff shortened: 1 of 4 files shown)


src/python/ensembl/io/genomio/fasta/compare.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+class CompareFasta:
+    """Read and compare the fasta sequences"""
+
+    def __init__(self, fasta1: Path, fasta2: Path, output_dir: str) -> None:
+        self.fasta1 = Path(fasta1)
+        self.fasta2 = Path(fasta2)
+        self.output_dir = Path(output_dir)
+        self.comp: List[str] = []
```
JAlvarezJarreta, line 81: This attribute may need some explanation.

src/python/ensembl/io/genomio/fasta/compare.py
```diff
@@ -0,0 +1,284 @@
...
+        # Compare number of sequences
+        if len(seq1) != len(seq2):
+            self.comp.append(
+                "WARNING: Different number of sequences: "
+                f"fasta1 [ n = {len(seq1)} ] -Vs- fasta2 [ n = {len(seq2)} ]"
+            )
+            logging.warning("Different number of sequences: fasta1 compared to fasta2")
+
+        common, group_comp = self.find_common_groups(seq1_dict, seq2_dict)
+        self.comp += group_comp
```
JAlvarezJarreta, line 103:
```suggestion
```
`group_comp` is already `self.comp` (you may want to change the function to only return `common`).

src/python/ensembl/io/genomio/fasta/compare.py
```diff
@@ -0,0 +1,284 @@
...
+                        f" ({group1.count} vs {group2.count}): {group1} and {group2}"
+                    )
+
+        return common, self.comp
+
+    def write_results(self) -> None:
+        """
+        Write the comparison results to a file in the output directory.
+
+        """
```
JAlvarezJarreta, line 202:
```suggestion
        """Write the comparison results to a file in the output directory."""
```

src/python/ensembl/io/genomio/fasta/compare.py
```diff
@@ -0,0 +1,284 @@
...
+        output_file = Path.joinpath(self.output_dir, "compare.log")
```
JAlvarezJarreta, line 203:
```suggestion
        output_file = self.output_dir / "compare.log"
```

src/python/ensembl/io/genomio/fasta/compare.py
Dishalodha: yes makes sense, thank you

src/python/ensembl/io/genomio/fasta/compare.py
```diff
@@ -229,15 +221,15 @@ def compare_seq_for_Ns(self, only1: dict[str, SeqGroup], only2: dict[str, SeqGro
                 seq2_N = seq_2.count("N")
 
                 if abs(seq1_N - seq2_N) == abs(len1 - len2):
-                    self.comp.append(f"Please check extra Ns added in your fasta2 in {name1} and {name2}")
+                    self.comp.append(f"Please check extra Ns added in your fasta_core in {name1} and {name2}")
```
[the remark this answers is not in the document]
Dishalodha: yes you are right, both files are interchangeable currently, I am not sure maybe based on how we use it we could standardise it later. I think in terms of output here it would matter it helps to point which file has extra Ns in terms of readability in report?

PR #480, opened 2025-12-05
microbes valid_structures additions

manuelcarbajo:
This PR adds some valid_structures that were found necessary during the last upload of FungiDB genomes by the Microbes team.
The genomes were successfully loaded after adding these modifications.


Dishalodha (approved):
Looks good !

PR #486, opened 2026-03-03
JSON schema for loading of repeat and ncRNA features

markquintontulloch:
This schema aims to capture all the information for loading repeat / ncRNA features into coreDBs / GenomeDB.  Each JSON file would relate to a single analysis (e.g. RepeatMasker) and the features identified by that analysis.

@JAlvarezJarreta  What are your thoughts about loading analysis metadata that should be (relatively) constant, e.g. analysis description.  Would that be loaded along with the repeat features for that analysis, or in a prior load?

This schema assumes that this is not already present prior to loading the features.  This has the advantage that all metadata relating to the analysis is present in each JSON file, and only a single load is required for each analysis and its set of features.  However, it also means that all metadata has to be present in each JSON file produced from chunked output prior to merging (and the merging code needs to check for consistency across all metadata fields).

Personally, I think the advantages of keeping everything together outweigh the disadvantages.

### DIFF
--- a/src/python/ensembl/io/genomio/data/schemas/load_features.json
+++ b/src/python/ensembl/io/genomio/data/schemas/load_features.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "https://.../.../src/python/ensembl/io/genomio/data/schemas/load_features.json",
-  "title": "Genomio repeat & ncRNA feature load schema",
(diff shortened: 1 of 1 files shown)


JAlvarezJarreta (approved):
Great addition, thank you.

The only thing that is present in this document and I'm not so sure of is the presence of meta keys - these are evolving from main to beta, and even in beta they may change, so keeping them here is only asking for added maintenance. But that is just my opinion.

I have also covered your question (or at least part of it) in a comment.

PR #487, opened 2026-03-07
Refactor: Improve error handling, prevent dictionary mutation, and fix type hints in OSID organism script

RohitKattimani:
While going through the OSID tools, I noticed a couple of spots where the script could malfunction or behave unexpectedly, particularly during batch processing.

### The Fixes:
* **Dictionary Mutation:** In `update_organism`, the script was deleting the `organismName` key from the original input dictionary. If that data is needed elsewhere in a larger workflow, it’s gone. I switched this to use a local copy and `.pop()` to keep the original data intact.
* **Non-JSON Server Errors:** Currently, if the OSID server hits a snag (like a 500 error) and returns HTML instead of JSON, the script crashes while trying to print the error. I’ve wrapped the JSON parsing in a try/except so it falls back to printing the raw response text instead of blowing up.
* **Network Failures:** Added a `try/except` around the `requests` calls to handle timeouts/connection issues more gracefully.
* **Type Hinting:** Caught a missing `str` hint on the `user` param in `add_or_update_organism` and added it for consistency.

### Testing:
Ran this locally with some mock responses:
1. Confirmed that the input dictionary is no longer mutated after an update.
2. Verified that HTML error pages from the server now print as raw text rather than triggering a `JSONDecodeError`.

### DIFF
--- a/scripts/stable_id_alloc/osid_add_organisms.py
+++ b/scripts/stable_id_alloc/osid_add_organisms.py
@@ -71,7 +71,7 @@ def add_organism(url: str, user: str, key: str, org_data: Dict[str, str]) -> Non
             print(f"Error: {result.status_code}")
             try:
                 print(json.dumps(result.json(), indent=4))
-            except ValueError:  # Catches JSONDecodeError if in case respobse is not JSON
+            except ValueError:  # Catches JSONDecodeError if in case response is not JSON
                 print(result.text)
     except requests.exceptions.RequestException as e:
         print(f"Network error occurred while adding organism: {e}")
--- a/scripts/stable_id_alloc/osid_add_organisms.py
+++ b/scripts/stable_id_alloc/osid_add_organisms.py
@@ -61,31 +61,48 @@ def add_organism(url: str, user: str, key: str, org_data: Dict[str, str]) -> Non
 
     add_org_url = url + "/organisms"
     headers = {"Content-type": "application/json", "Accept": "application/json"}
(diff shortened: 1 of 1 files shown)


scripts/stable_id_alloc/osid_add_organisms.py
```diff
@@ -61,31 +61,48 @@ def add_organism(url: str, user: str, key: str, org_data: Dict[str, str]) -> Non
...
+    try:
+        result = requests.post(add_org_url, auth=(user, key), headers=headers, json=org_data, timeout=10)
+        
+        if result and result.status_code == 200:
+            print("Successfully added organism")
+        else:
+            print(f"Error: {result.status_code}")
+            try:
+                print(json.dumps(result.json(), indent=4))
+            except ValueError:  # Catches JSONDecodeError if in case respobse is not JSON
```
JAlvarezJarreta:
```suggestion
            except ValueError:  # Catches JSONDecodeError if in case response is not JSON
```

JAlvarezJarreta (changes requested):
Thank you for submitting these updates. There are two things to amend:
- One small typo
- `print(result.text)` is not a good alternative as our system expects a JSON as output, so your catch alternative would not work

RohitKattimani:
Thanks for the review. I've addressed the typo and updated the exception blocks to wrap non-JSON server responses into a valid JSON object. This ensures the script's output remains parsable by the pipeline even during the edge cases.

JAlvarezJarreta (approved):
Look good to me. Thank you!

PR #491, opened 2026-05-27
Convert TRF / RepeatMasker output to JSON

markquintontulloch:
Creates JSON according to `load_features` schema from RepeatMasker .out files and TRF .dat files.  To be expanded for other repeat/ncRNA feature tools as new pipelines are developed.


src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+"""Constructs a GenomIO JSON document from output of feature identification tools."""
+
+import argparse
+import hashlib
+import json
+import logging
+import re
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from pathlib import Path
```
JAlvarezJarreta, line 23:
```suggestion
from dataclasses import dataclass
from datetime import datetime, timezone
import hashlib
import json
import logging
from pathlib import Path
import re
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+
+
+from Bio import SeqIO
+
+import ensembl.io.genomio
+
```
JAlvarezJarreta:
```suggestion
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+        """
+        norm_name = self.name.strip()
+        norm_class = self.repeat_class.strip()
+        norm_type = self.repeat_type.strip()
+        norm_seq = "".join(self.seq.split()).upper()
+        payload = f"{norm_name}\t{norm_class}\t{norm_type}\t{norm_seq}".encode("utf-8")
+        return hashlib.sha256(payload).hexdigest()
+
+
+DEFAULT_ANALYSIS_PARAMS = {
```
JAlvarezJarreta: Usually constant go above any code, so the `Consensus` class should go below these constants block.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+    Returns:
+        int: Parsed integer value.
+
+    Raises:
+        ValueError: If the token is not a valid integer.
+    """
+    try:
+        return int(token)
+    except ValueError:
+        raise ValueError(f"Invalid {field_name!r} in {path}: token={token!r}, line={raw_line!r}")
```
JAlvarezJarreta: Why are you using `!r` on string parameters? To my understanding it should not make a difference...

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+    Returns:
+        float: Parsed floating-point value.
+
+    Raises:
+        ValueError: If the token is not a valid float.
+    """
+    try:
+        return float(token)
+    except ValueError:
+        raise ValueError(f"Invalid {field_name!r} in {path}: token={token!r}, line={raw_line!r}")
```
JAlvarezJarreta: Not sure if worth it, but you could merge this function and the one above into a single one by passing the class you want to cast the token to:

```suggestion
def _parse_token(class_type: Type, token: str, field_name: str, raw_line: str, path: Path) -> Any:
    """
    Parses a field from TRF output into the specified class type.

    Args:
        class_type: Class type to cast the field to.
        token: Raw token to parse.
        field_name: Field name used in error messages.
        raw_line: Original input line.
        path: Input file path.

    Returns:
        Parsed value.

    Raises:
        ValueError: If the token cannot be casted into the provided type.
    """
    try:
        return typing.cast(class_type, token)
    except ValueError:
        raise ValueError(
                f"Invalid {field_name!r} in {path}: type={class_type}, token={token!r}, line={raw_line!r}"
        )
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+
+
+def _file_last_modified_time(file_path: Path) -> str:
```
JAlvarezJarreta, line 124:
```suggestion
def _file_last_modified_time(file_path: Path) -> str:
    """Returns the last modified time of the given file."""
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+
+    Args:
+        seq_region_start: Start coordinate on the sequence region.
+        seq_region_end: End coordinate on the sequence region.
+        repeat_start: Start coordinate on the repeat consensus.
+        repeat_end: End coordinate on the repeat consensus.
+        line: Original input line for error reporting.
+
+    Returns:
+        bool: `True` if all coordinates are valid, `False` otherwise.
```
JAlvarezJarreta:
```suggestion
        `True` if all coordinates are valid, `False` otherwise.
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+    Parses an integer field from TRF output.
+
+    Args:
+        token: Raw token to parse.
+        field_name: Field name used in error messages.
+        raw_line: Original input line.
+        path: Input file path.
+
+    Returns:
+        int: Parsed integer value.
```
JAlvarezJarreta:
```suggestion
        Parsed integer value.
```
Return type is only required if you don't have type hints, otherwise it is best to remove duplication.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+    Parses a RepeatMasker consensus library FASTA file into a dictionary of Consensus records.
+
+    The parser expects FASTA headers in the format:
+        >consensus_name#repeat_class/repeat_type
+
+    Args:
+        rm_consensus_lib_path: Path to the RepeatMasker consensus library FASTA file.
+
+    Returns:
+        tuple[dict[tuple[str, str, str], str], dict[str, Consensus]]: A tuple containing:
```
JAlvarezJarreta:
```suggestion
        A tuple containing:
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+        logging.warning(
+            f"repeat_end < repeat_start in {input_path}: "
+            f"repeat_start={repeat_start}, repeat_end={repeat_end}, line={line!r}"
+        )
+        no_warnings = False
+
+    return no_warnings
+
+
+def _parse_rm_consensus_library(
```
JAlvarezJarreta: I'm slightly uneasy with the use of `rm` to refer to RepeatMasker as `rm` has a very distinct meaning in Linux/programming environments... Maybe replace in the function's name for the complete name and then do not refer to it in the variables? e.g. `rm_consensus_lib_path` -> `consensus_library_path`.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+                repeat_type=_map_rm_repeat_type(repeat_type),
+                seq=str(record.seq),
+            )
+            consensus_key = consensus_obj.sha256_key()
+            consensus_keys_by_triplet[(repeat_name, repeat_class, repeat_type)] = consensus_key
+            consensuses_by_key[consensus_key] = consensus_obj
+    return consensus_keys_by_triplet, consensuses_by_key
+
+
+def parse_repeatmasker_out(
```
JAlvarezJarreta:
```suggestion
def parse_repeatmasker_out_file(
```
Or `output`, but if you are not familiar with the `.out` extension of RepeatMasker this makes it more meaningful.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+) -> tuple[list[dict], dict[str, Consensus]]:
+    """
+    Parse a RepeatMasker .out file into repeat feature dictionaries and consensus records.
+
+    If a RepeatMasker consensus library FASTA file is provided, consensus sequences will be
+    extracted from the library and associated with features based on their repeat names.
+    Otherwise, consensus records will be created with empty sequences.
+
+    Returns:
+        tuple[list[dict], dict[str, Consensus]]: A tuple containing:
```
JAlvarezJarreta:
```suggestion
        A tuple containing:
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+            - repeat feature dictionaries
+            - repeat consensus dictionary keyed by consensus SHA256 digest
+
+    Raises:
+        ValueError: If the RepeatMasker output contains malformed rows or invalid
+            coordinate values.
```
JAlvarezJarreta:
```suggestion
        ValueError: If the RepeatMasker output contains malformed rows or invalid coordinate values.
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+def parse_repeatmasker_out(
+    input_path: Path, rm_consensus_lib_path: Path | None
+) -> tuple[list[dict], dict[str, Consensus]]:
+    """
+    Parse a RepeatMasker .out file into repeat feature dictionaries and consensus records.
+
+    If a RepeatMasker consensus library FASTA file is provided, consensus sequences will be
+    extracted from the library and associated with features based on their repeat names.
+    Otherwise, consensus records will be created with empty sequences.
+
```
JAlvarezJarreta, line 257: `Args:` block missing.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+        for raw_line in fh:
+            line = raw_line.strip()
+            if not line:
+                continue
+
+            lower_line = line.lower()
+            if "no repetitive sequences detected" in lower_line:
+                break
+            if "only contains ambiguous bases" in lower_line:
+                break
```
JAlvarezJarreta:
```suggestion
            if lower_line in {"no repetitive sequences detected", "only contains ambiguous bases"}:
                break
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+
+            if line.startswith("SW") or line.startswith("score") or line.startswith("There were"):
```
JAlvarezJarreta: Have just learned about this lovely trick:
```suggestion
            if line.startswith(("SW", "score", "There were"))
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+                continue
+
+            columns = line.split()
+            if columns[-1] == "*":
+                columns.pop()
+
+            if len(columns) < 14 or len(columns) > 15:
+                raise ValueError(
+                    f"Expected 14 or 15 columns in {input_path}, got {len(columns)}: line={line!r}"
+                )
```
JAlvarezJarreta: Would it make sense in this case to alter this method to check the whole file in one go before reporting the issues? One thing I was thinking about (and Claude backs up) is to report the detailed line issue in the log file, and just raise a "generic" exception message after processing the file if something went wrong. But this depends on how large the file is (if files are meant to be very long, a threshold may be worth adding that stops execution if more than, let's say, 100 lines are incorrect.

Just a thought/idea though, I'm happy to leave things as they are if you think it is not worth it.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+            if consensus_triplet in consensus_keys_by_triplet:
+                consensus_key = consensus_keys_by_triplet[consensus_triplet]
+                feature["repeat_consensus"] = consensus_key
+
+            features.append(feature)
+
+    return features, consensuses_by_key
+
+
+def parse_trf_out(input_path: Path) -> tuple[list[dict], dict[str, Consensus]]:
```
JAlvarezJarreta:
```suggestion
def parse_trf_dat_file(input_path: Path) -> tuple[list[dict], dict[str, Consensus]]:
```
To follow the same logic as above.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+    """
+    Parses a TRF .dat file into repeat feature dictionaries and consensus records.
+
+    Coordinates are converted from TRF's sequence-relative coordinates to genomic
+    coordinates if the header provides a window like:
+        Sequence: NC_135497.1:379500001-380000000
+
+    Returns:
+        tuple[list[dict], dict[str, Consensus]]: A tuple containing:
```
JAlvarezJarreta:
```suggestion
        A tuple containing:
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+            c_pct = _parse_float(columns[9], "c_pct", line, input_path)
+            g_pct = _parse_float(columns[10], "g_pct", line, input_path)
+            t_pct = _parse_float(columns[11], "t_pct", line, input_path)
+            entropy = _parse_float(columns[12], "entropy", line, input_path)
+            motif = columns[13] if len(columns) >= 14 else ""
+
+            if seq_region is None:
+                raise ValueError(
+                    f"TRF sequence header not found before data lines in {input_path}: line={line!r}"
+                )
```
JAlvarezJarreta: This should be moved above, to line 449, because if the header is missing it may be best to not process the data lines, right?

Another thought, as I'm not familiar with the content of this file: would you need to set `seq_region = None` in any other line case to avoid processing data lines as belonging to the "wrong" seq_region?

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+    analysis_display_label: str,
+    analysis_description: str,
+    program: str,
+    program_version: str,
+    program_parameters: str | None,
+    source_provider: str,
+    is_primary: bool,
+) -> None:
+    """
+    Creates a GenomIO JSON document from feature identificationtool output.
```
JAlvarezJarreta:
```suggestion
    Creates a GenomIO JSON document from feature identification tool output.
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+
+    Args:
+        input: Path to the input file containing repeat masking results (e.g. RepeatMasker .out file).
+        output: Path to the output JSON file to be created.
+        rm_consensus_lib: Optional path to a FASTA file containing consensus sequences for the
+            RepeatMasker library used.
```
JAlvarezJarreta: Do we want to set this one to `None` by default?

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+        source_provider: Name of the source provider for the features (e.g. "Ensembl").
+        is_primary: Whether these features are primary or secondary annotations.
+    """
+    features: list[dict[str, object]] = []
+    consensuses_by_key: dict[str, Consensus] = {}
+
+    if analysis_logic_name.startswith("repeatmask_"):
+        features, consensuses_by_key = parse_repeatmasker_out(input_path, rm_consensus_lib_path)
+    elif analysis_logic_name == "trf":
+        features, consensuses_by_key = parse_trf_out(input_path)
```
JAlvarezJarreta: If nothing else, at least I would log the "`analysis_logic_name` not recognised" scenario.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+    output_path.write_text(json.dumps(json_doc, indent=2) + "\n", encoding="utf-8")
+
+
+def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
+    """
+    Parse command-line arguments for JSON conversion.
+
+    Args:
+        argv: Optional list of command-line arguments. If `None`, arguments are
+            taken from `sys.argv`.
```
JAlvarezJarreta:
```suggestion
        argv: Optional list of command-line arguments. If `None`, arguments are taken from ``sys.argv``.
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+
+    Returns:
+        argparse.Namespace: Parsed command-line arguments.
```
JAlvarezJarreta:
```suggestion
        Parsed command-line arguments.
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+        help="Parameters supplied to the program used for feature identification.",
+    )
+    parser.add_argument("--source-provider", default="Ensembl")
+    parser.add_argument("--is-primary", action="store_true")
+    parser.add_log_arguments()
+    parser.add_argument("--version", action="version", version=ensembl.io.genomio.__version__)
+
+    args = parser.parse_args(argv)
+    if args.analysis_logic_name not in DEFAULT_ANALYSIS_PARAMS:
+        raise ValueError(f"Unsupported analysis logic name: {args.analysis_logic_name}")
```
JAlvarezJarreta: Instead of this check, I would add `choices=DEFAULT_ANALYSIS_PARAMS.keys()` to `--analysis-logic-name` to make the parser handle this (it will also add these to the `--help` message, which is quite convenient).

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+DEFAULT_ANALYSIS_PARAMS = {
+    "repeatmask_customlib": {
+        "description": (
+            'Repeats identified by <a rel="external" href="http://www.repeatmasker.org">RepeatMasker</a>, '
+            "using a custom library of <em>ab initio</em> repeat profiles for this species."
+        ),
+        "display_label": "Repeats: Custom library",
+        "program": "RepeatMasker",
+    },
+    "repeatmask_rmlib": {
```
JAlvarezJarreta: I assume these are the default names we have had, because to me naming these `repeatmasker_custom_lib` and `repeatmasker_repbase_lib` makes much more sense.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+    parser.add_log_arguments()
+    parser.add_argument("--version", action="version", version=ensembl.io.genomio.__version__)
+
+    args = parser.parse_args(argv)
+    if args.analysis_logic_name not in DEFAULT_ANALYSIS_PARAMS:
+        raise ValueError(f"Unsupported analysis logic name: {args.analysis_logic_name}")
+    if hasattr(args, "rm_consensus_lib") and not args.analysis_logic_name.startswith("repeatmask_"):
+        raise ValueError(
+            f"--rm-consensus-lib is not applicable for analysis logic name: {args.analysis_logic_name}"
+        )
```
JAlvarezJarreta: Whilst this is fine, I find it cleaner to use subparsers in this case for `trf` _vs_ `repeatmasker` (and for the latter have the `--consensus-lib` as well as `--repeats-lib` with `choices=['custom', 'repbase']`). So everything is much under controlled dictionaries to avoid having to check many things. Or we could do 2 levels of subparsers for `repeatmasker` if you want to include the default values for `analysis_display_label`, `analysis_description` and `program` instead of the conditional checks you have below.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,725 @@
...
+    return args
+
+
+def main(argv: list[str] | None = None) -> None:
+    """
+    Run the JSON conversion command-line entry point.
+
+    Args:
+        argv: Optional list of command-line arguments. If `None`, arguments are
+            taken from ``sys.argv``.
```
JAlvarezJarreta:
```suggestion
        argv: Optional list of command-line arguments. If `None`, arguments are taken from ``sys.argv``.
```

JAlvarezJarreta (changes requested):
_Partial review: Since some of my suggestions may change some of the logic of your module, I'm going to wait to review the test once we are happy with the module itself._

Nice and clear module 👍

markquintontulloch:
@JAlvarezJarreta I think all comments should have been addressed now.  Errors are now collated before raising and subparsers used for CLI arguments.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -113,58 +57,72 @@ def sha256_key(self) -> str:
...
+        return parser(token)
     except ValueError:
         raise ValueError(f"Invalid {field_name!r} in {path}: token={token!r}, line={raw_line!r}")
 
 
+def _format_parse_errors(parser_name: str, input_path: Path, errors: list[str]) -> str:
+    """Formats multiple parser errors into a single exception message."""
+    return f"Found {len(errors)} errors while parsing {parser_name} in {input_path}:\n" + "\n".join(
+        f"- {error}" for error in errors
+    )
```
JAlvarezJarreta, line 121:
```suggestion
    error_list = "\n".join(f"- {error}" for error in errors)
    return f"Found {len(errors)} errors while parsing {parser_name} in {input_path}:\n{error_list}"
```
I find this a little bit more readable, but just a preference.

JAlvarezJarreta (commented):
Great update, I'm very happy with the changes on reporting analysis of complete files and the use of subparsers: it looks much clearner and useful. 👍

Going to dive into the unit tests now.

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+        seq: Consensus sequence.
+
+    Returns:
+        str:Expected SHA-256 digest.
+    """
+    norm_name = name.strip()
+    norm_class = repeat_class.strip()
+    norm_type = repeat_type.strip()
+    norm_seq = "".join(seq.split()).upper()
+    payload = f"{norm_name}\t{norm_class}\t{norm_type}\t{norm_seq}".encode("utf-8")
```
JAlvarezJarreta, line 51:
```suggestion
    payload = f"{name}\t{repeat_class}\t{repeat_type}\t{seq}".encode("utf-8")
```
I would avoid duplicating code on the class method, and "force" data from tests to conform the expected output: this is the best way I can think to ensure test data mistakes are captured.

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+            repeat_type="Unknown",
+            seq="ccgg",
+        ),
+    }
+
+    expected_keys_by_triplet = {
+        ("AluY", "SINE", "Alu"): _sha256_key("AluY", "SINE", "Unknown", "acgt"),
+        ("Foo", "LINE", "Unknown"): _sha256_key("Foo", "LINE", "Unknown", "AATT"),
+        ("Bar", "Unknown", "Unknown"): _sha256_key("Bar", "Unknown", "Unknown", "ccgg"),
+    }
```
JAlvarezJarreta:
```suggestion
@pytest.mark.parametrize(
    ("entries"),
    [
        [
            ("AluY", "SINE", "Unknown", "ACGT"),
            ("Foo", "LINE", "Unknown", "AATT"),
            ("Bar", "Unknown", "Unknown", "CCGG")
        ]
    ],
)
def test_parse_repeatmasker_consensus_library(data_dir: Path, entries: list[tuple]) -> None:
    """
    Tests the ``convert_to_genomio_json._parse_repeatmasker_consensus_library()`` internal helper.

    Args:
        data_dir: Module's test data directory fixture.
    """
    fasta_path = data_dir / "repeatmodeler" / "classified_mixed.fa"

    expected_consensuses = {
        _sha256_key(*x): convert_to_genomio_json.Consensus(
            name=x[0], repeat_class=x[1], repeat_type=x[2], seq=x[3]
        )
        for x in entries
    }

    expected_keys_by_triplet = {(x[0], x[1], x[2]): _sha256_key(*x) for x in entries}
```
Alternative approach relying on the change suggested above for `_sha256_key()` and making the code more adaptable to future changes.

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+        ),
+    ],
+)
+def test_parse_repeatmasker_output_success(
+    data_dir: Path,
+    filename: str,
+    expected_features: list[dict[str, object]],
+) -> None:
+    """
+    Tests successful parsing of RepeatMasker ``.out`` files.
```
JAlvarezJarreta:
```suggestion
    Tests successful parsing of RepeatMasker `.out` files.
```

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    Args:
+        data_dir: Module's test data directory fixture.
+        filename: Input RepeatMasker ``.out`` filename.
```
JAlvarezJarreta:
```suggestion
        filename: Input RepeatMasker `.out` filename.
```

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+        param(
+            "repeatmasker_out/error_c_repeat_start_invalid.out",
+            r"Invalid 'repeat_start'",
+            id="Rejects invalid reverse strand repeat start",
+        ),
+    ],
+)
+def test_parse_repeatmasker_output_errors(data_dir: Path, filename: str, error_pattern: str) -> None:
+    """
+    Tests that malformed RepeatMasker ``.out`` rows raise errors.
```
JAlvarezJarreta:
```suggestion
    Tests that malformed RepeatMasker `.out` rows raise errors.
```

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    Args:
+        data_dir: Module's test data directory fixture.
+        filename: Name of RepeatMasker ``.out`` file containing invalid data.
```
JAlvarezJarreta:
```suggestion
        filename: Name of RepeatMasker `.out` file containing invalid data.
```

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+    filename: str,
+    warning_pattern: str,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    """
+    Tests that ``convert_to_genomio_json.parse_repeatmasker_output()`` logs a warning and skips invalid records.
+
+    Args:
+        data_dir: Module's test data directory fixture.
+        filename: Input RepeatMasker ``.out`` filename containing invalid coordinates.
```
JAlvarezJarreta:
```suggestion
        filename: Input RepeatMasker `.out` filename containing invalid coordinates.
```

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+    ],
+)
+def test_parse_trf_output_success(
+    data_dir: Path,
+    filename: str,
+    expected_features: list[dict[str, object]],
+    expected_consensuses: dict[str, convert_to_genomio_json.Consensus],
+) -> None:
+    """
+    Tests successful parsing of TRF ``.dat`` examples.
```
JAlvarezJarreta:
```suggestion
    Tests successful parsing of TRF `.dat` examples.
```

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    Args:
+        data_dir: Module's test data directory fixture.
+        filename: Input TRF ``.dat`` filename.
```
JAlvarezJarreta:
```suggestion
        filename: Input TRF `.dat` filename.
```

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+                "TRF sequence header not found before data lines",
+                "unprocessed_entries=2",
+                "first_data_line=7",
+                "not:Invalid 'start'",
+            ],
+            id="TRF missing Sequence line before EOF",
+        ),
+    ],
+)
+def test_parse_output_collates_all_errors(
```
JAlvarezJarreta, line 755: Not the standard approach, but tests what we want to happy with this.

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+                    "g_pct": 25.0,
+                    "t_pct": 25.0,
+                    "trf_parameters": "2 7 7 80 10 50 500",
+                },
+            },
+            id="Tandem Repeat Finder",
+        ),
+    ],
+)
+def test_create_genomio_json(
```
JAlvarezJarreta: Do we want to patch the parse methods to avoid running them? Also to isolate errors in each function's code.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,777 @@
...
+    repeatmasker_parser = subparsers.add_parser(
+        "repeatmasker", help="Convert RepeatMasker output to GenomIO JSON."
+    )
+    repeatmasker_parser.set_defaults(program="RepeatMasker")
+    repeatmasker_subparsers = repeatmasker_parser.add_subparsers(dest="repeatmasker_mode", required=True)
+
+    def _add_repeatmasker_common_args(subparser: ArgumentParser) -> None:
+        _add_common_arguments(subparser)
+        subparser.add_argument_src_path(
+            "--rm-consensus-lib",
```
JAlvarezJarreta:
```suggestion
            "--consensus-lib",
```
Since you are already in the repeatmasker subparser, the context should be clear enough.

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+            str(input),
+            "--output",
+            str(output),
+            "--program-version",
+            "4.1.5",
+        ]
+    )
+
+    with pytest.raises(SystemExit):
+        convert_to_genomio_json.parse_args(argv)
```
JAlvarezJarreta: I would test unexpected/wrong argument values, but this is testing that the subparser was not assigned the "wrong" arguments, which is something that should be already checked above when asserting what was returned by the parser.

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+    ],
+)
+def test_main_passes_expected_arguments(
+    mock_create_genomio_json: Mock,
+    data_dir: Path,
+    tmp_path: Path,
+    use_repeatmodeler_lib: bool,
+) -> None:
+    """
+    Tests ``convert_to_genomio_json.main()`` passes parsed arguments through to `create_genomio_json()` using real input files.
```
JAlvarezJarreta, line 1094:
```suggestion
    Tests ``convert_to_genomio_json.main()`` passes parsed arguments through to
    ``convert_to_genomio_json.create_genomio_json()`` using real input files.
```

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    Args:
+        mock_create_genomio_json: Mock for `create_genomio_json()`.
```
JAlvarezJarreta:
```suggestion
        mock_create_genomio_json: Mock for ``convert_to_genomio_json.create_genomio_json()``.
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -0,0 +1,777 @@
...
+
+    # TRF subparser
+    trf_parser = subparsers.add_parser("trf", help="Convert TRF output to GenomIO JSON.")
+    _add_common_arguments(trf_parser)
+    trf_parser.set_defaults(
+        analysis_logic_name="trf",
+        analysis_display_label="Tandem repeats (TRF)",
+        analysis_description=(
+            '<a rel="external" href="https://tandem.bu.edu/trf/trf.html">Tandem Repeats Finder</a> '
+            "locates adjacent copies of a pattern of nucleotides."
```
JAlvarezJarreta, line 697: Most likely I missed it: Is this used elsewhere that having an html embedded string is needed? If not this is going to look most likely unreadable on the terminal.

JAlvarezJarreta (changes requested):
Tests look good. A few suggestions/comments.

Moreover, I know they are silly functions, but it would be safer to have a test for `_parse_token()`, `_format_parse_errors()`, `_file_last_modified_time()` and `_has_valid_parsed_coordinates()` just to ensure expected behaviour is preserved.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
markquintontulloch: This is just what is currently stored in the `analysis_description.description` table of the core DBs

markquintontulloch:
@JAlvarezJarreta I think all comments should have been addressed now

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
JAlvarezJarreta: Let's put a pin on this to revisit when moving to the new architecture then 👍

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -70,51 +75,213 @@ def test_consensus_sha256_key_normalises_fields() -> None:
...
+    seq_region_start: int,
+    seq_region_end: int,
+    repeat_start: int,
+    repeat_end: int,
+    expectation: ContextManager,
+    warning_pattern: str | None,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    """
+    Tests ``convert_to_genomio_json._has_valid_parsed_coordinates()`` correctly validatescoordinates.
```
JAlvarezJarreta:
```suggestion
    Tests ``convert_to_genomio_json._has_valid_parsed_coordinates()`` correctly validates coordinates.
```

JAlvarezJarreta (approved):
One tiny typo, but everything looks great. Cheers!

PR #495, opened 2026-06-08
Python CICD fixes

markquintontulloch:


markquintontulloch:
@JAlvarezJarreta This fixes the python stuff ~~but the `perl:critic` job is failing now due to being unable to download dependencies for some reason~~.  Well that seems to be working now too!

src/python/ensembl/io/genomio/fasta/recombine.py
```diff
@@ -104,8 +105,8 @@ def _parse_fasta_headers(file_path: StrPath) -> Iterator[tuple[str, str]]:
         Tuples of (record_id, description_without_id).
     """
     with open_gz_file(file_path) as fh:
-        for record in SeqIO.parse(fh, "fasta"):
-            yield record.id, seq_description_without_id(record)
+        headers = [(record.id, seq_description_without_id(record)) for record in SeqIO.parse(fh, "fasta")]
+    yield from headers
```
JAlvarezJarreta, line 109: Out of curiosity, do you know the rationale behind this change? The original code looked good to me.

src/python/ensembl/io/genomio/features/combine_json.py
```diff
@@ -399,6 +409,7 @@ def _combine_feature_docs(
     documents: Iterable[tuple[Path, dict[str, JsonValue]]],
     feature_list_key: str,
     coerce_feature: CoerceFeatureFunction,
+    *,
```
JAlvarezJarreta, line 412: Why is this needed? As in, what raised the need to add this now?

src/python/tests/features/test_combine_json.py
```diff
@@ -12,6 +12,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# pylint: disable=too-many-lines
```
JAlvarezJarreta, line 16: We had it below to file's docstring in other places, but I quite like it here more 👍

cicd/gitlab/parts/python.gitlab-ci.yml
```diff
@@ -70,6 +70,8 @@ python:pylint:src:
 
 python:pylint:tests:
   extends: .python:pylint
+  variables:
+    PYLINT_CMD: pylint --rcfile pyproject.toml --disable=protected-access
```
JAlvarezJarreta: I'm not an advocate of particular configurations only applied to the CI pipeline because when you check manually you are likely to forget adding this additional configuration. From what I can see, the best is to create a `.pylintrc` file inside the `tests` folder with the following content:
```
[MESSAGES CONTROL]
disable=protected-access
```

src/python/ensembl/io/genomio/fasta/split.py
```diff
@@ -207,26 +212,27 @@ def write_record(
...
 
     def close(self) -> None:
+        """Closes any open FASTA and AGP file handles."""
         self._fh.close()
         if self._agp_fh:
             self._agp_fh.close()
             self._agp_fh = None
 
 
-def _check_contents_deletable(dir: Path, output_file_re: re.Pattern[str]) -> None:
```
JAlvarezJarreta, line 222: I missed that one, good catch!

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -91,6 +93,22 @@ def sha256_key(self) -> str:
...
+@dataclass(frozen=True)
+class RepeatMaskerParsedRow:
+    """Parsed RepeatMasker row and its consensus lookup triplet."""
+
+    feature: dict[str, object]
+    consensus_triplet: tuple[str, str, str]
+
+
+@dataclass(frozen=True)
+class TrfParsedRow:
```
JAlvarezJarreta: Not sure if TRF is common to be seen as Trf, but if so:

```suggestion
class TRFParsedRow:
```

src/python/tests/features/test_convert_to_genomio_json.py
```diff
@@ -288,6 +287,134 @@ def test_parse_repeatmasker_consensus_library(
...
+        expected: Expected repeat class and raw repeat type.
+    """
+    assert (
+        convert_to_genomio_json._parse_repeatmasker_repeat_class_field(
+            Path("input.out"),
+            repeat_class_field,
+            "raw line",
+        )
+        == expected
+    )
```
JAlvarezJarreta: I would opt to have these in 2 parts, as I believe it should help slightly with readability:

```suggestion
    output = convert_to_genomio_json._parse_repeatmasker_repeat_class_field(
        Path("input.out"), repeat_class_field, "raw line"
    )
    assert output == expected
```

JAlvarezJarreta (changes requested):
Thank you very much for taking care of fixing this! 🎉 

After the updates on `convert_to_genomio_json` I'm wondering if we should not have opted for 3 modules instead of 1: the conversion module, one for RepeatMasker parser and one for TRF parser. It is more modular y definitely more structured and easier to expand in the future. Maybe create a JIRA ticket for this to be done in the future?

src/python/ensembl/io/genomio/features/combine_json.py
markquintontulloch: Too many positional arguments

src/python/ensembl/io/genomio/fasta/recombine.py
markquintontulloch: pylint didn't like it:
W0135: The context used in function '_parse_fasta_headers' will not be exited. (contextmanager-generator-missing-cleanup)

cicd/gitlab/parts/python.gitlab-ci.yml
markquintontulloch: I can't find a way to get this to work in such a way that the same command can be run from the command line for test/non-test cases since the --rcfile needs to be specified (apart from putting the disable annotation in all test files, which I'd rather avoid)

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
markquintontulloch: updated

src/python/tests/features/test_convert_to_genomio_json.py
markquintontulloch: Agreed, updated

cicd/gitlab/parts/python.gitlab-ci.yml
JAlvarezJarreta: So we either move all the pylint configuration to two separate `.pylintrc` files (replacing the config part in `pyproject.toml`) or we follow your suggestion (although I would then drop the variable `PYLINT_CMD` since it is basically not being "reused"). I will leave the final decision up to you.

JAlvarezJarreta (approved):

PR #497, opened 2026-06-15
Fix TRF splitting

markquintontulloch:
`ls` fails if splitting results in very long lists of files, using `find` instead


vsitnik (approved):
all good. thank you!

PR #498, opened 2026-06-16
Ruff linting and formatting fixes

markquintontulloch:


src/python/ensembl/io/genomio/events/dump.py
```diff
@@ -39,47 +39,50 @@
...
         return not (self.has_old_id() or self.has_new_id())
 
 
-class UnsupportedEvent(ValueError):
-    """If an event is not supported"""
+class UnsupportedEventError(ValueError):
+    """If an event is not supported."""
+
+
+UnsupportedEvent = UnsupportedEventError  # Backwards compatibility
```
JAlvarezJarreta: Good thing to do, but I would personally remove this backwards compatibility in this case (I doubt any repository out there will depend on this code).

src/python/ensembl/io/genomio/events/format.py
```diff
@@ -33,33 +32,34 @@ class IdsMapper:
...
+
         """
         mapping = {}
         with map_file.open("r") as map_fh:
             for line in map_fh:
                 if line == "":
                     continue
                 items = line.split("\t")
-                if len(items) < 2:
+                if len(items) < 2:  # noqa: PLR2004
```
JAlvarezJarreta, line 48: Should we add this as an ignored check in `pyproject.toml` instead?

src/python/ensembl/io/genomio/events/load.py
```diff
@@ -46,6 +46,7 @@ class IdEvent:
     release_date: str
 
     def __str__(self) -> str:
+        """Return a string representation of this event."""
```
JAlvarezJarreta, line 49: I thought magic methods didn't need a docstring given their nature, but good to know.

src/python/ensembl/io/genomio/fasta/chunk.py
```diff
@@ -58,34 +59,38 @@ def check_chunk_size_and_tolerance(
     chunk_tolerance: int,
     error_f: Callable[[str], None] = _on_value_error,
 ) -> None:
-    """Check the chunk size and the tolerance are positive and chunk size is not too small
+    """Check the chunk size and the tolerance are positive and chunk size is not too small.
 
     Args:
         chunk_size: Chunk size to check
         chunk_tolerance: Chunk tolerance to check
+        error_f: A callable to call with an error message if checks fail, default is to raise ValueError.
```
JAlvarezJarreta:
```suggestion
        error_f: A callable to call with an error message if checks fail, default is to raise `ValueError`.
```

src/python/ensembl/io/genomio/fasta/process.py
```diff
@@ -56,16 +53,18 @@ def get_peptides_to_exclude(genbank_path: PathLike, seqr_to_exclude: Set[str]) -
...
+    genbank_infile: PathLike | None,
     fasta_outfile: PathLike,
     peptide_mode: bool = False,
 ) -> None:
-    """
+    """Prepare FASTA data.
+
     Args:
-        fasta_file: Input FASTA file - DNA / Protein
+        fasta_infile: Input FASTA file - DNA / Protein
```
JAlvarezJarreta:
```suggestion
        fasta_infile: Input FASTA file - DNA / Protein.
```

src/python/ensembl/io/genomio/fasta/split.py
```diff
@@ -196,49 +207,53 @@ def write_record(
...
         self.record_count += 1
         self.file_len += len(sequence)
 
         if self.write_agp:
-            assert agp_object_id is not None, "AGP object ID must be provided if writing AGP entries"
-            assert agp_start is not None, "AGP start must be provided if writing AGP entries"
-            assert agp_end is not None, "AGP end must be provided if writing AGP entries"
-            assert agp_part_nr is not None, "AGP part no. must be provided if writing AGP entries"
+            if agp_object_id is None:
+                raise AssertionError("AGP object ID must be provided if writing AGP entries")
```
JAlvarezJarreta, line 221: I wonder why ruff didn't like the previous code: to me it is quite standard and readable...

src/python/ensembl/io/genomio/genbank/download.py
```diff
@@ -58,7 +59,7 @@ def download_genbank(accession: str, output_file: PathLike) -> None:
     entrez_params["id"] = accession
     logging.debug(f"Getting file from {entrez_url} with params {entrez_params}")
     result = requests.get(entrez_url, params=entrez_params, timeout=60)
-    if result and result.status_code == 200:
+    if result and result.status_code == HTTP_OK_STATUS:
```
JAlvarezJarreta: I found about this the other day, which is quite cool IMO:
```python
    response.raise_for_status()
```

Not sure how we want to put it in place though, this is the suggestion from the documentation:
```python
try:
    response = requests.get('https://api.example.com/data')
    response.raise_for_status()  # Raises an error for bad responses
except requests.exceptions.HTTPError as err:
    print(f"HTTP error occurred: {err}")
except Exception as err:
    print(f"An error occurred: {err}")
```

src/python/ensembl/io/genomio/features/combine_json.py
```diff
@@ -653,8 +653,12 @@ def combine_feature_json(
...
 
     if mismatched:
-        paths = "\n".join(f"  - {p}" for p in mismatched[:20])
-        extra = "" if len(mismatched) <= 20 else f"\n  (+{len(mismatched) - 20} more)"
+        paths = "\n".join(f"  - {p}" for p in mismatched[:_MISMATCH_ERRORS_TO_REPORT])
+        extra = (
+            ""
+            if len(mismatched) <= _MISMATCH_ERRORS_TO_REPORT
+            else f"\n  (+{len(mismatched) - _MISMATCH_ERRORS_TO_REPORT} more)"
+        )
```
JAlvarezJarreta:
```suggestion
        if len(mismatched) <= _MISMATCH_ERRORS_TO_REPORT:
            extra = ""
        else:
            extra = f"\n  (+{len(mismatched) - _MISMATCH_ERRORS_TO_REPORT} more)"
```
1 less line of code, much more readable IMO.

src/python/ensembl/io/genomio/features/convert_to_genomio_json.py
```diff
@@ -313,15 +315,16 @@ def _parse_repeatmasker_strand_coordinates(
 
     Raises:
         ValueError: If the strand token or coordinate tokens are invalid.
+
     """
     strand_token = columns[8]
-    if strand_token == "+":
+    if strand_token == "+":  # noqa: S105
```
JAlvarezJarreta: We do not handle passwords in our code so I would add this to the list of codes to ignore in `pyproject.toml`.

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -51,14 +60,15 @@ class GBParseError(Exception):
     """Error when parsing the Genbank file."""
 
 
-class UnsupportedData(Exception):
+class UnsupportedDataError(Exception):
     """When an expected data is not supported by the current parser."""
 
 
+UnsupportedData = UnsupportedDataError  # Alias for backward compatibility
```
JAlvarezJarreta: Same comment as before.

src/python/ensembl/io/genomio/genbank/extract_data.py
```diff
@@ -71,12 +81,9 @@ def __init__(self, out_dir: PathLike) -> None:
...
 
 class FormattedFilesGenerator:
-    """
-    Contains a parser to load data from a file, and output a set of files that follow our schema
-    for input into a core database
-    """
+    """Loads data from a file, and outputs a set of files following our core database input schema."""
 
-    locations = {
+    locations: ClassVar[dict[str, str]] = {
```
JAlvarezJarreta, line 82: Interesting, didn't know about this at all 🤓

src/python/ensembl/io/genomio/genome_metadata/dump.py
```diff
@@ -132,16 +133,10 @@ def filter_genome_meta(
         if key in genome_metadata:
             filtered_metadata[key] = {}
             for subkey, value_type in subfilter.items():
-                if isinstance(value_type, str):
-                    value_type = type(value_type)
-                if isinstance(value_type, int):
-                    value_type = type(value_type)
+                converter = type(value_type) if isinstance(value_type, (str, int)) else value_type
```
JAlvarezJarreta, line 136: Another useful thing I didn't know about 🤓

src/python/ensembl/io/genomio/genome_metadata/dump.py
```diff
@@ -244,11 +244,12 @@ def convert_dict(meta_dict: dict) -> dict:
...
     db_url: URL, input_filter: StrPath | None, meta_update: bool, append_db: bool
 ) -> dict[str, Any]:
-    """Setup main stages of genome meta dump from user input arguments provided.
+    """Prepare main stages of genome meta dump from user input arguments provided.
+
     Args:
         db_url: Target core database URL.
         input_filter: Input JSON containing subset of meta table values to filter on.
-        no_update: Deactivate additional meta updating.
+        meta_update: Deactivate additional meta updating.
```
JAlvarezJarreta: If the definition is correct, this is a very bad variable name...

src/python/ensembl/io/genomio/genome_metadata/prepare.py
```diff
@@ -153,18 +154,16 @@ def add_species_metadata(genome_metadata: Dict, ncbi_data: Dict) -> None:
     if "organism_name" in organism:
         species.setdefault("scientific_name", organism["organism_name"])
 
-    try:
+    with contextlib.suppress(KeyError):
```
JAlvarezJarreta, line 157: Another cool trick 🤓

src/python/ensembl/io/genomio/gff3/id_allocator.py
```diff
@@ -14,20 +14,24 @@
...
 
+MAX_IDS_PER_STABLE_BASE = 10
 
-class InvalidStableID(ValueError):
+
+class InvalidStableIDError(ValueError):
     """Raised when there is a problem with an stable ID."""
 
 
+InvalidStableID = InvalidStableIDError  # Alias for backward compatibility
```
JAlvarezJarreta: Same comment as before.

src/python/ensembl/io/genomio/gff3/id_allocator.py
```diff
@@ -14,20 +14,24 @@
...
+__all__ = ["InvalidStableID", "InvalidStableIDError", "StableIDAllocator"]
 
 from dataclasses import dataclass, field
 import logging
 import re
-from typing import Dict, List, Optional, Set
 
 from .features import GFFSeqFeature
 
+MAX_IDS_PER_STABLE_BASE = 10
```
JAlvarezJarreta, line 25: We can undo this change if we decide to ignore the warning that flagged this change.

src/python/ensembl/io/genomio/gff3/simplifier.py
```diff
@@ -63,10 +61,11 @@ def from_gff(self, in_gff_path: PathLike, excluded: Optional[List[str]] = None)
                 self.append(clean_record)
 
     def to_gff(self, out_gff_path: PathLike) -> None:
-        """Writes the current list of records in a GFF3 file.
+        """Writethe current list of records in a GFF3 file.
```
JAlvarezJarreta:
```suggestion
        """Write the current list of records in a GFF3 file.
```

src/python/ensembl/io/genomio/manifest/check_integrity.py
```diff
@@ -213,24 +202,24 @@ def check_lengths(
...
         Args:
             list1: dict containing length and id of the sequence from fasta files.
             list2: dict containing length and id in the retrieved from the gff.
             name:  string
-
-        allowed_len_diff : None to to not accept differences in length between list1 and list2.
-            The value can be changed based on how much difference in sequence length we are wanting to accept.
-
-        special_diff: set as False when no special length difference is expected between the lists.
+            allowed_len_diff : None to to not accept differences in length between list1 and list2.
```
JAlvarezJarreta:
```suggestion
            allowed_len_diff : ``None`` to not accept differences in length between `list1` and `list2`.
```

src/python/ensembl/io/genomio/manifest/check_integrity.py
```diff
@@ -213,24 +202,24 @@ def check_lengths(
...
+                    The value can be changed based on how much difference in sequence length we are
+                    wanting to accept.
+            special_diff: set as False when no special length difference is expected between the lists.
```
JAlvarezJarreta:
```suggestion
            special_diff: Set as ``False`` when no special length difference is expected between the lists.
```

src/python/ensembl/io/genomio/manifest/manifest.py
```diff
@@ -20,19 +20,22 @@
...
+
 class ManifestError(Exception):
     """Could not load a manifest file."""
 
 
 class Manifest:
     """Records of a manifest file and its files and md5 checksums."""
 
-    _same_names = {
+    _same_names: ClassVar[set[str]] = {
```
JAlvarezJarreta, line 38: Out of curiosity, could you not just do `set[str]` here?

src/python/ensembl/io/genomio/manifest/manifest_stats.py
```diff
@@ -293,31 +297,33 @@ def _retrieve_gff_gene_lengths(
...
                     peps[pep_id] = pep_length
                 all_peps[pep_id] = pep_length
 
     def load_agp_seq_regions(self, agp_dict: dict | None) -> None:
-        """AGP files describe the assembly of larger sequence objects using smaller objects.
+        """Load sequence regions from AGP files.
 
-        E.g. describes the assembly of scaffolds from contigs.
+        AGP files describe the assembly of larger sequence objects using smaller objects.
+        e.g. describes the assembly of scaffolds from contigs.
```
JAlvarezJarreta:
```suggestion
        E.g. describes the assembly of scaffolds from contigs.
```

src/python/ensembl/io/genomio/seq_region/exceptions.py
```diff
@@ -16,8 +16,12 @@
...
+    "UnknownMetadataError",
 ]
 
 
-class UnknownMetadata(Exception):
+class UnknownMetadataError(Exception):
     """If a metadata is not supported or recognized."""
+
+
+UnknownMetadata = UnknownMetadataError  # Backwards compatibility
```
JAlvarezJarreta: Same as above.

src/python/ensembl/io/genomio/seq_region/gbff.py
```diff
@@ -83,10 +83,10 @@ def get_organelle(self, molecule_location: Mapping[str, str] = MOLECULE_LOCATION
...
                 location = molecule_location[organelle]
             except KeyError as exc:
-                raise UnknownMetadata(f"Unrecognized sequence location: {organelle}") from exc
+                raise UnknownMetadataError(f"Unrecognized sequence location: {organelle}") from exc
             break
         return location
 
     def is_circular(self) -> bool:
-        """Returns True if the record says that the sequence is circular, False otherwise."""
+        """Return `True` if the record says that the sequence is circular, `False` otherwise."""
```
JAlvarezJarreta:
```suggestion
        """Return ``True`` if the record says that the sequence is circular, ``False`` otherwise."""
```
I have finally dove into this topic (one "`" vs 2), and it seems it should always be 2 but when we add `:func:` (or similar), in those cases it is only 1.

src/python/tests/assembly/test_download.py
```diff
@@ -409,24 +412,29 @@ def test_retrieve_assembly_data(
...
-        expectation: Context manager expected raise exception
-    """
+        mock_retrieve: Mock of ``ensembl.io.genomio.assembly.download.md5_files()``.
+        mock_download_single_file: Mock of ``ensembl.io.genomio.assembly.download._download_file()``.
+        mock_download_files: Mock of ``ensembl.io.genomio.assembly.download.download_files()``.
+        mock_file_select: Mock of ``ensembl.io.genomio.assembly.download.get_files_selection()``.
+        mock_ftp: Mock of ``ensembl.io.genomio.assembly.download.FTP``.
+        data_dir: Module's test data directory fixture.
+        accession: Accession of desired genome assembly.
+        is_dir: Param to define state of result output dir.
```
JAlvarezJarreta:
```suggestion
        is_dir: Defines state of result output directory.
```

src/python/tests/assembly/test_status.py
```diff
@@ -209,10 +211,7 @@ def test_singularity_image_setter(
     else:
         expected_cache_path = Path()
     # Get expected container URL used to pull the container
-    if datasets_version:
-        expected_container_url = datasets_version
-    else:
-        expected_container_url = DATASETS_SINGULARITY["datasets_version_url"]
+    expected_container_url = datasets_version or DATASETS_SINGULARITY["datasets_version_url"]
```
JAlvarezJarreta, line 214: I find this way of coding this behaviour very confusing, as my brain tends to think a boolean value is assigned to the variable. Proposing an alternative, but leave the final call to you:

```suggestion
    expected_container_url = (
        datasets_version
        if datasets_version
        else DATASETS_SINGULARITY["datasets_version_url"]
    )
```

.coveragerc
```diff
@@ -0,0 +1,3 @@
+[report]
+exclude_also =
+    if TYPE_CHECKING:
```
JAlvarezJarreta: Cannot this be added to the `tool.coverage.report` block in `pyproject.toml`?

src/python/tests/fasta/test_split.py
```diff
@@ -180,47 +187,75 @@ def test_init(
...
+                fasta_file=Path("in.fa"),
+                out_dir=tmp_path,
+                write_agp=write_agp,
+                unique_file_names=False,
             )
             writer.close()
 
-    def test_open_new_file(self, tmp_path: Path) -> None:
+        with pytest.raises(RuntimeError, match=rf"{exc_msg}"):
+            create_and_close()
```
JAlvarezJarreta: Why have you moved the code that was inside this `with` to its own function? I do not see the logic/need (but I bet there is one).

src/python/tests/gff3/test_restructure.py
```diff
@@ -35,58 +34,58 @@ class FeatGenerator:
     region = "LOREM"
     source = "Foo"
 
-    def make(self, ftype: str, number: int = 1) -> List[GFFSeqFeature]:
-        """Returns a list with a defined number of features of a given type."""
+    def make(self, ftype: str, number: int = 1) -> list[GFFSeqFeature]:
```
JAlvarezJarreta, line 37: _[Not to be done in this PR]_ This function needs to be changed to return just one `GFFSeqFeature` as the iterative part/list object returned is never used in this test file.

JAlvarezJarreta (changes requested):
Great work! A few comments and suggestions.

src/python/ensembl/io/genomio/events/format.py
markquintontulloch: I feel like this is more of a case-by-case thing: sometimes the rule makes sense, other times (like here) it's overly strict.

src/python/ensembl/io/genomio/manifest/manifest.py
markquintontulloch: RUF012 (mutable-class-default) wants it marked as a class attribute to indicate it's shared across all instances

src/python/tests/fasta/test_split.py
markquintontulloch: PT012 (pytest-raises-with-multiple-statements) doesn't like multiple statements inside a `pytest.raises()` block

src/python/tests/assembly/test_status.py
markquintontulloch: black and ruff don't like the proposed change

markquintontulloch:
@JAlvarezJarreta Unless commented above, I think all comments have been addressed

src/python/ensembl/io/genomio/genbank/download.py
```diff
@@ -55,16 +53,26 @@ def download_genbank(accession: str, output_file: PathLike) -> None:
...
+            timeout=60,
+        )
+        response.raise_for_status()
+
         with Path(output_file).open("wb") as gbff:
-            gbff.write(result.content)
-        logging.info(f"GenBank file written to {output_file}")
-        return
-    raise DownloadError(f"Could not download the genbank ({accession}) file: {result}")
+            gbff.write(response.content)
```
JAlvarezJarreta: The writing into the file should be outside the `try-except` block so if it raises an exception for some reason it is not silently ignored.

JAlvarezJarreta (changes requested):
One tiny change, and we should be good to go.

src/python/tests/assembly/test_status.py
JAlvarezJarreta: It is one of those things I will need to live with 🤷‍♂️ thank you for trying!

src/python/tests/fasta/test_split.py
JAlvarezJarreta: Then we could just leave the `writer.close()` outside the context manager and it would be ok, right? Trying to preserve the original code as much as possible and avoid creating new function objects when they do not bring any clarity to the code.

markquintontulloch: In this case I think we just need to remove the `writer.close()` altogether as it's not what is being tested here

src/python/ensembl/io/genomio/genbank/download.py
markquintontulloch: Moved

markquintontulloch: Corresponding test needed updating too

src/python/ensembl/io/genomio/fasta/split.py
markquintontulloch: https://docs.astral.sh/ruff/rules/assert-false/

JAlvarezJarreta (approved):
We got there! 🎉

PR #500, opened 2026-06-19
Update documentation framework

JAlvarezJarreta:
Migration from mkdocs + material to Sphinx + PyData. Additional changes:

- Updated the documentation generation and move it to GitHub Actions (it was not working, so this should fix it), including the coverage report (so the badge should be visible now)
- Move the license header check to GitHub Actions
- Updated NOTICE year update workflow to enhance security
- Dropped all the documentation that has nothing to do with the GenomIO Python library (if we want to keep it, we may need to think of an alternative way to do that, but it was significantly outdated)
- Add missing library dependency: SQLAlchemy
- Add Sphinx extension to generate a entry point table from `pyproject.toml` content (generated by Claude and reviewed by me)


JAlvarezJarreta:
Conflicts resolved, and new changes included in PR #504.

PR #502, opened 2026-06-23
Fix logic name from RepeatMasker Repbase analysis

markquintontulloch:
Make logic name consistent with current core DB analysis logic name for RepeatMasker Repbase.  Doesn't affect module logic, only JSON output.


JAlvarezJarreta (approved):
Looks good to me 👍

PR #503, opened 2026-06-29
ENSGENOMIO-25 Create modularised JSON conversion package

markquintontulloch:


src/python/ensembl/io/genomio/features/convert_to_genomio_json/args.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Returns:
+        Parsed command-line arguments.
+
+    """
+    parser = ArgumentParser(description="Constructs a GenomIO JSON document from feature output.")
+    parser.add_argument("--version", action="version", version=ensembl.io.genomio.__version__)
+
+    subparsers = parser.add_subparsers(dest="tool", required=True)
+    _add_trf_parser(subparsers)
```
vsitnik: don't we want to toy with `@classmethod`s. We can create a structure to point to the derived classes.
The derived classes could register themselves and the base class could iterate over that structure and call corresponding methods. We'd be able to store all the functionality related to a specific tool in one class if we want to.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/document.py
```diff
@@ -0,0 +1,106 @@
...
+from pathlib import Path
+
+from ensembl.io.genomio.features.convert_to_genomio_json.base import Consensus, file_last_modified_time
+from ensembl.io.genomio.features.convert_to_genomio_json.repeatmasker import parse_repeatmasker_output
+from ensembl.io.genomio.features.convert_to_genomio_json.trf import parse_trf_output
+
+__all__ = ["create_genomio_json"]
+
+
+def create_genomio_json(  # noqa: PLR0913
```
vsitnik: could we mark ruff controlling comments some how?
will it allow to have a prefix of some kind?
i.e. `# noqa: PLR0913  -- ruff linter option`

src/python/ensembl/io/genomio/features/convert_to_genomio_json/repeatmasker.py
```diff
@@ -0,0 +1,329 @@
...
+                repeat_type=map_repeatmasker_repeat_consensus_type(repeat_class),
+                seq=str(record.seq),
+            )
+            consensus_key = consensus_obj.sha256_key()
+            consensus_keys_by_triplet[(repeat_name, repeat_class, repeat_type)] = consensus_key
+            consensuses_by_key[consensus_key] = consensus_obj
+    return consensus_keys_by_triplet, consensuses_by_key
+
+
+def parse_repeatmasker_repeat_class_field(
```
vsitnik: do we need `repeatmasker` infix? Do we need to preserve it once we have classes in different files? Can we turn it into a class/static method?

src/python/ensembl/io/genomio/features/convert_to_genomio_json/trf.py
```diff
@@ -0,0 +1,321 @@
...
+        ValueError: If the row is malformed or contains invalid coordinates.
+
+    """
+    columns = line.split()
+    if len(columns) < MIN_TRF_COLUMNS:
+        raise ValueError(
+            f"Expected at least {MIN_TRF_COLUMNS} columns in {input_path}, got {len(columns)}: line={line!r}"
+        )
+
+    start = parse_token(int, columns[0], "start", line, input_path)
```
vsitnik, line 177: can we somehow use pydantic for validation?

vsitnik (dismissed):
cool. thanks!

src/python/ensembl/io/genomio/features/convert_to_genomio_json/trf.py
markquintontulloch: I'm not convinced that pydantic is a better approach here.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/args.py
markquintontulloch: I've refactored to use `@classmethod`s in tool-specific converter classes

src/python/ensembl/io/genomio/features/convert_to_genomio_json/document.py
markquintontulloch: yes, done

src/python/ensembl/io/genomio/features/convert_to_genomio_json/repeatmasker.py
markquintontulloch: methods renamed to remove redundancy, but I don't see the value in converting these to static methods

markquintontulloch:
@vsitnik - there has been significant change since your review, would you mind re-reviewing?  Thanks

src/python/ensembl/io/genomio/features/convert_to_genomio_json/registry.py
```diff
@@ -0,0 +1,38 @@
...
+    RepeatMaskerRepbaseConverter,
+)
+from ensembl.io.genomio.features.convert_to_genomio_json.trf import TrfConverter
+
+__all__ = [
+    "CONVERTERS_BY_LOGIC_NAME",
+    "TOP_LEVEL_CONVERTERS",
+]
+
+CONVERTERS_BY_LOGIC_NAME = {
```
vsitnik: cool. thanks. mb we can add some automation here, but on a later stage.

Though, should we have something like
```
TOP_LEVEL_CONVERTERS = [...]
CONVERTERS_BY_LOGIC_NAME = {
 c.analysis_logic_name : c for c in TOP_LEVEL_CONVERTERS
}
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -0,0 +1,155 @@
...
+@dataclass(frozen=True)
+class Consensus:
+    """Repeat consensus record used for feature-to-consensus linking."""
+
+    name: str
+    repeat_class: str
+    repeat_type: str
+    seq: str
+
+    def sha256_key(self) -> str:
```
vsitnik, line 65: cool

src/python/ensembl/io/genomio/features/convert_to_genomio_json/trf.py
vsitnik: we could keep constraints/validation logic more or less in one place
https://pydantic.dev/docs/validation/latest/concepts/validators/#model-validators
Though, I suggest, I can try this with the dust stuff first

vsitnik (approved):
cool. thanks.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/converters.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+ParseFeaturesResult = tuple[list[dict[str, object]], dict[str, Consensus]]
+
+
+@dataclass(frozen=True)
+class ConverterOptions:
+    """Tool-specific options supplied to feature converters."""
+
+    repeatmasker_consensus_lib_path: Path | None = None
```
Dishalodha: Do we intend to expand this function to add more tools later?

src/python/ensembl/io/genomio/features/convert_to_genomio_json/document.py
```diff
@@ -0,0 +1,103 @@
...
+        "program_version": program_version,
+    }
+    if program_parameters is not None:
+        analysis["program_parameters"] = program_parameters
+
+    json_doc: dict[str, object] = {
+        "analysis": analysis,
+        "source": {
+            "source_provider": source_provider,
+            "is_primary": bool(is_primary),
```
Dishalodha: Do we need to add another bool here if the type is already specified as bool?

src/python/ensembl/io/genomio/features/convert_to_genomio_json/registry.py
```diff
@@ -0,0 +1,38 @@
...
+    RepeatMaskerCustomConverter,
+    RepeatMaskerRepbaseConverter,
+)
+from ensembl.io.genomio.features.convert_to_genomio_json.trf import TrfConverter
+
+__all__ = [
+    "CONVERTERS_BY_LOGIC_NAME",
+    "TOP_LEVEL_CONVERTERS",
+]
+
```
Dishalodha: Could we add a short note explaining the distinction between TOP_LEVEL_CONVERTERS and CONVERTERS_BY_LOGIC_NAME?

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -0,0 +1,155 @@
...
+        raise ValueError(
+            f"seq_region_end < seq_region_start in {input_path}: "
+            f"start={seq_region_start}, end={seq_region_end}, line={line!r}"
+        )
+
+    no_warnings = True
+    if repeat_start < 1 or repeat_end < 1:
+        logging.warning(
+            f"Invalid repeat coordinates in {input_path}: "
+            f"repeat_start={repeat_start}, repeat_end={repeat_end}, line={line!r}"
```
Dishalodha: Instead of logging should we consider removing the negative or short repeats? 
Also consider logging into another file? I just wonder if we run this on nextflow it would be easy to miss the warnings.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/repeatmasker.py
```diff
@@ -0,0 +1,418 @@
...
+    """Converter for RepeatMasker output generated with Repbase."""
+
+    analysis_logic_name = "repeatmask_repbase"
+
+    @classmethod
+    def add_parser(cls, subparsers: argparse._SubParsersAction) -> None:
+        """Add the Repbase RepeatMasker mode parser."""
+        repbase_parser = subparsers.add_parser(
+            "repbase",
+            help="Convert RepeatMasker output geneterated using Repbase.",
```
Dishalodha:
```suggestion
            help="Convert RepeatMasker output generated using Repbase.",
```

src/python/ensembl/io/genomio/features/convert_to_genomio_json/registry.py
markquintontulloch: There's a difference between what's in TOP_LEVEL_CONVERTERS and what we want as keys for CONVERTERS_BY_LOGIC NAME

src/python/ensembl/io/genomio/features/convert_to_genomio_json/converters.py
markquintontulloch: Yes, or at least possibly (I don't know whether other tools will need tool-specific options)

src/python/ensembl/io/genomio/features/convert_to_genomio_json/document.py
markquintontulloch: no, removed

src/python/ensembl/io/genomio/features/convert_to_genomio_json/registry.py
markquintontulloch: Comments added

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: Yes, entries that generate warnings should be being skipped.  This was being done for RepeatMasker but not TRF - I've fixed that now.

Good point about the warnings getting lost - I'll have a think about the best way to address that.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/repeatmasker.py
markquintontulloch: fixed

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
Dishalodha: On taking another look, if they are skipped maybe we don't need to log them! Warning should be fine.

markquintontulloch: I've refactored so that repeat coordinate validation failures raise exceptions (like the seq region coordinates) instead of logging warnings.  Exceptions for each file are accumulated and reported together.  If this causes issues (because repeat coordinate errors are occurring and we would want to ignore them but accept the rest of the file) then we can revert to warnings and come up with a better way of reporting.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -0,0 +1,147 @@
...
+
+T = TypeVar("T")
+
+__all__ = [
+    "Consensus",
+    "file_last_modified_time",
+    "format_parse_errors",
+    "parse_token",
+    "validate_parsed_coordinates",
+]
```
JAlvarezJarreta, line 32: [PEP8] Module level “dunders” go before the import block.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -0,0 +1,147 @@
...
+
+    """
+    try:
+        return parser(token)
+    except ValueError as exc:
+        raise ValueError(f"Invalid {field_name!r} in {path}: token={token!r}, line={raw_line!r}") from exc
+
+
+def format_parse_errors(parser_name: str, input_path: Path, errors: list[str]) -> str:
+    """Format multiple parser errors into a single exception message."""
```
JAlvarezJarreta, line 203: `Args` and `Returns` blocks missing.

src/python/tests/features/test_convert_to_genomio_json_base.py
```diff
@@ -0,0 +1,192 @@
...
+        assert base.parse_token(parser, token, field_name, raw_line, Path("input.out")) == expected
+
+
+def test_format_parse_errors() -> None:
+    """Test ``base.format_parse_errors()`` formats a counted bullet list."""
+    assert base.format_parse_errors(
+        "TRF output",
+        Path("input.dat"),
+        ["first error", "second error"],
+    ) == ("Found 2 errors while parsing TRF output in input.dat:\n- first error\n- second error")
```
JAlvarezJarreta:
```suggestion
    output = base.format_parse_errors("TRF output", Path("input.dat"), ["first error", "second error"])
    assert output == ("Found 2 errors while parsing TRF output in input.dat:\n- first error\n- second error")
```
I find it more readable this way.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/registry.py
```diff
@@ -0,0 +1,45 @@
...
+    RepeatMaskerConverter,
+    RepeatMaskerCustomConverter,
+    RepeatMaskerRepbaseConverter,
+)
+from ensembl.io.genomio.features.convert_to_genomio_json.trf import TrfConverter
+
+__all__ = [
+    "CONVERTERS_BY_LOGIC_NAME",
+    "TOP_LEVEL_CONVERTERS",
+]
```
JAlvarezJarreta: Same as before re. module-level "dunders".

src/python/ensembl/io/genomio/features/convert_to_genomio_json/document.py
```diff
@@ -0,0 +1,103 @@
...
+"""Create GenomIO JSON documents from parsed repeat feature records."""
+
+import json
+from pathlib import Path
+
+from ensembl.io.genomio.features.convert_to_genomio_json.base import file_last_modified_time
+from ensembl.io.genomio.features.convert_to_genomio_json.converters import ConverterOptions
+from ensembl.io.genomio.features.convert_to_genomio_json.registry import CONVERTERS_BY_LOGIC_NAME
+
+__all__ = ["create_genomio_json"]
```
JAlvarezJarreta: Same as before.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/converters.py
```diff
@@ -0,0 +1,91 @@
...
+from pathlib import Path
+
+from ensembl.io.genomio.features.convert_to_genomio_json.base import Consensus
+from ensembl.utils.argparse import ArgumentParser
+
+__all__ = [
+    "ConverterOptions",
+    "FeatureConverter",
+    "ParseFeaturesResult",
+]
```
JAlvarezJarreta: Same as before.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/converters.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+ParseFeaturesResult = tuple[list[dict[str, object]], dict[str, Consensus]]
```
JAlvarezJarreta: Add a `"""Description"""` below to be included in the documentation.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/converters.py
JAlvarezJarreta: This is one of those cases where this is not following the intention of the modularisation: the class name looks generic, but then you have added only one very specific tool element. Either you define it as an abstract class that needs to be implemented where needed, or you don't define it at all (and document its need properly).

src/python/ensembl/io/genomio/features/convert_to_genomio_json/args.py
```diff
@@ -0,0 +1,48 @@
...
+
+import argparse
+
+import ensembl.io.genomio
+from ensembl.io.genomio.features.convert_to_genomio_json.registry import TOP_LEVEL_CONVERTERS
+
+from ensembl.utils.argparse import ArgumentParser
+from ensembl.utils.logging import init_logging_with_args
+
+__all__ = ["parse_args"]
```
JAlvarezJarreta: [PEP8] Module level “dunders” go before the import block.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/repeatmasker.py
```diff
@@ -0,0 +1,414 @@
...
+]
+
+REPEATMASKER_COMPILED_MAPPINGS = [(re.compile(pattern), mapped) for pattern, mapped in REPEATMASKER_MAPPINGS]
+
+__all__ = [
+    "RepeatMaskerConverter",
+    "RepeatMaskerCustomConverter",
+    "RepeatMaskerParsedRow",
+    "RepeatMaskerRepbaseConverter",
+]
```
JAlvarezJarreta: Same re. module-level "dunders".

src/python/ensembl/io/genomio/features/convert_to_genomio_json/trf.py
```diff
@@ -0,0 +1,357 @@
...
+MIN_TRF_COLUMNS = 13
+
+TRF_SEQUENCE_RE = re.compile(r"^Sequence:\s+(?P<seq_region>\S+?)(?::(?P<start>\d+)-\d+)?\s*$")
+
+TRF_PARAMETERS_RE = re.compile(r"^Parameters:\s+(?P<params>.+)\s*$")
+
+__all__ = [
+    "TRFParsedRow",
+    "TrfConverter",
+]
```
JAlvarezJarreta: Same re. module-level "dunders".

src/python/ensembl/io/genomio/features/convert_to_genomio_json/repeatmasker.py
```diff
@@ -0,0 +1,414 @@
...
+        cls,
+        input_path: Path,
+        options: ConverterOptions | None = None,
+    ) -> ParseFeaturesResult:
+        """Parse custom-library RepeatMasker output."""
+        options = options or ConverterOptions()
+        return parse_output(input_path, options.repeatmasker_consensus_lib_path)
+
+
+class RepeatMaskerRepbaseConverter(RepeatMaskerCustomConverter):
```
JAlvarezJarreta: It is risky to make one converter based on the other to save duplicating 10 lines of code. I would base it on `FeatureConverter`.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/converters.py
```diff
@@ -0,0 +1,91 @@
...
+    def parse_features(
+        cls,
+        input_path: Path,
+        options: ConverterOptions | None = None,
+    ) -> ParseFeaturesResult:
+        """Parse features and consensus records from a tool output file."""
+        raise NotImplementedError
+
+
+def _add_common_arguments(subparser: ArgumentParser) -> None:
```
JAlvarezJarreta: This method should be part of `FeatureConverter`.

JAlvarezJarreta (changes requested):
_Partial review_

markquintontulloch:
@JAlvarezJarreta I think that everything we have discussed (and the above comments) has been addressed in the latest commits

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -58,6 +78,98 @@ def sha256_key(self) -> str:
...
+    def add_parser(cls, subparsers: argparse._SubParsersAction) -> None:
+        """Add this converter's CLI parser."""
+        raise NotImplementedError
+
+    @staticmethod
+    def add_common_arguments(subparser: ArgumentParser) -> None:
+        """Add arguments shared by all supported analysis subcommands."""
+        subparser.add_argument_src_path(
+            "--input",
+            metavar="IN",
```
JAlvarezJarreta: I personally like the default metavar added by this method more (`PATH`) because it is a little bit more descriptive than `IN`. But just a preference.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -58,6 +78,98 @@ def sha256_key(self) -> str:
...
+        )
+        subparser.add_argument(
+            "--program-parameters",
+            default=argparse.SUPPRESS,
+            help="Parameters supplied to the program used for feature identification.",
+        )
+        subparser.add_argument(
+            "--source-provider",
+            default="Ensembl",
+            help="Source provider for the features (default: Ensembl).",
```
JAlvarezJarreta:
```suggestion
           help="Source provider for the features.",
```
The default value is already shown by the `--help` option in the ensembl.utils Argparse.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -58,6 +78,98 @@ def sha256_key(self) -> str:
...
+
+
+class FeatureConverter:
+    """Base class for feature output converters."""
+
+    analysis_logic_name: str | None = None
+    analysis_display_label: str | None = None
+    analysis_description: str | None = None
+    command: str | None = None
+    program: str | None = None
```
JAlvarezJarreta, line 124: Since this is not an abstract class or a dataclass, I feel like the `__init__()` is missing. The documentation around these classes and how to add a new tool-specific submodule needs to be included as well.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -145,3 +257,163 @@ def validate_parsed_coordinates(
...
+# Register only converters that should appear as first-level CLI tool commands
+# under the main parser. A top-level converter may parse records itself, or it
+# may only create nested mode-specific subcommands.
+TOP_LEVEL_CONVERTERS: list[type[Any]] = []
+
+
+def register_converter(converter: type[FeatureConverter]) -> type[FeatureConverter]:
+    """Register a concrete converter by analysis logic name."""
+    if not converter.analysis_logic_name:
+        raise ValueError(f"Converter {converter.__name__} has no analysis logic name")
```
JAlvarezJarreta, line 320: Following my `__init__()` comment above, do we want to create the constructor to force a FeatureConverter class to always set an `analysis_logic_name` to avoid having to check this? (and any other conditions)

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -145,3 +257,163 @@ def validate_parsed_coordinates(
...
+    registered_converter = CONVERTERS_BY_LOGIC_NAME.get(converter.analysis_logic_name)
+    if registered_converter is not None and registered_converter is not converter:
```
JAlvarezJarreta: You may want to create a comparison function for the FeatureConverter class, because `registered_converter is not converter` will evaluate to false if `converter` is a different object even if the attributes have exactly the same values as `registered_converter`. Or use `deepdiff`.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -145,3 +257,163 @@ def validate_parsed_coordinates(
...
+    converter_command = getattr(converter, "command", None)
+    if not converter_command:
+        raise ValueError(f"Top-level converter {converter.__name__} has no command")
+    for registered_converter in TOP_LEVEL_CONVERTERS:
+        registered_command = getattr(registered_converter, "command", None)
+        if registered_converter is not converter and registered_command == converter_command:
+            raise ValueError(
+                f"Top-level converter command {converter_command!r} already registered "
+                f"by {registered_converter.__name__}"
+            )
```
JAlvarezJarreta: Same as before regarding the comparison.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -145,3 +257,163 @@ def validate_parsed_coordinates(
...
+    """
+    parser = ArgumentParser(description="Constructs a GenomIO JSON document from feature output.")
+    parser.add_argument("--version", action="version", version=ensembl.io.genomio.__version__)
+
+    subparsers = parser.add_subparsers(dest="tool", required=True)
+    for converter in TOP_LEVEL_CONVERTERS:
+        converter.add_parser(subparsers)
+
+    args = parser.parse_args(argv)
+    init_logging_with_args(args)
```
JAlvarezJarreta, line 377: Why not adding `parser.add_log_arguments()` here instead of in `add_common_arguments()`?

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -145,3 +257,163 @@ def validate_parsed_coordinates(
...
+        raise ValueError(f"Top-level converter {converter.__name__} has no command")
+    for registered_converter in TOP_LEVEL_CONVERTERS:
+        registered_command = getattr(registered_converter, "command", None)
+        if registered_converter is not converter and registered_command == converter_command:
+            raise ValueError(
+                f"Top-level converter command {converter_command!r} already registered "
+                f"by {registered_converter.__name__}"
+            )
+    if converter not in TOP_LEVEL_CONVERTERS:
+        TOP_LEVEL_CONVERTERS.append(converter)
```
JAlvarezJarreta:
```suggestion
    TOP_LEVEL_CONVERTERS.append(converter)
```
Unless I missed something, at this point you already know the converter is not in this dictionary.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -145,3 +257,163 @@ def validate_parsed_coordinates(
...
+# Converter registry
+
+# Register every concrete converter class that parses a single analysis output,
+# keyed by the analysis.logic_name written to the JSON document.
+CONVERTERS_BY_LOGIC_NAME: dict[str, type[FeatureConverter]] = {}
+
+# Register only converters that should appear as first-level CLI tool commands
+# under the main parser. A top-level converter may parse records itself, or it
+# may only create nested mode-specific subcommands.
+TOP_LEVEL_CONVERTERS: list[type[Any]] = []
```
JAlvarezJarreta: Uppercase variables refer to constants, but you are planning to change the content of these, so they should be lowercase.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -145,3 +257,163 @@ def validate_parsed_coordinates(
...
+
+    """
+    try:
+        converter = CONVERTERS_BY_LOGIC_NAME[config.analysis_logic_name]
+    except KeyError:
+        raise ValueError(f"Unsupported analysis logic name: {config.analysis_logic_name}") from None
+    features, consensuses_by_key = converter.parse_features(config.input_path, config.converter_options)
+
+    analysis: dict[str, str] = {
+        "run_date": file_last_modified_time(config.input_path),
```
JAlvarezJarreta: Silly distinction, but `run_date` to me sounds like the day you started the program, not when it finished, so it should be the creation time of the file.

src/python/tests/features/test_convert_to_genomio_json_base.py
```diff
@@ -190,3 +175,389 @@ def test_validate_parsed_coordinates(
...
+    """Test converter registration rejects conflicting classes for one logic name."""
+    monkeypatch.setattr(base, "CONVERTERS_BY_LOGIC_NAME", {})
+
+    class ConflictingConverter(_DummyConverter):
+        """Converter with the same logic name as the dummy converter."""
+
+    base.register_converter(_DummyConverter)
+
+    with pytest.raises(ValueError, match=r"Converter logic name 'dummy' already registered"):
+        base.register_converter(ConflictingConverter)
```
JAlvarezJarreta, line 257: Alternatively, you could create 2 objects of the same class:
```suggestion
    converter = _DummyConverter()
    base.register_converter(converter)
    duplicated_converter = _DummyConverter()
    with pytest.raises(ValueError, match=r"Converter logic name 'dummy' already registered"):
        base.register_converter(duplicated_converter)
````

File: src/python/tests/features/test_convert_to_genomio_json_base.py [hunk 35]
```diff
@@ -190,3 +175,389 @@ def test_validate_parsed_coordinates(
...
+            """Return fixed parser output for document assembly tests."""
+            return (
+                [{"seq_region": input_path.name}],
+                {},
+            )
+
+    base.register_top_level_converter(_DummyConverter)
+
+    with pytest.raises(ValueError, match=r"Top-level converter command 'dummy' already registered"):
+        base.register_top_level_converter(ConflictingTopLevelConverter)
```
JAlvarezJarreta, line 329: Same as before, i.e. use `converter = _DummyConverter()` and `duplicated_converter = _DummyConverter()`.

JAlvarezJarreta (changes requested):
A huge improvement from last design in my opinion, great job! A few suggestions/comments, but almost there.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: I believe that would change the CLI.  For example,
`features_combine_genomio_json --log-level INFO trf`
would work, but I prefer the current form of
`features_combine_genomio_json trf --log-level INFO`

JAlvarezJarreta: Ah, it does not allow for the latter? Interesting... In that case I agree with you 100%.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: Using default now

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: Updated

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: I believe this should be an abstract class, I've updated it to inherit from ABC and added the appropriate `@abstractmethod` decorators.  Let me know if you don't think this is the correct approach.

Documentation added to `docs/user_guide/converter_modules.md` with `FeatureConverter` docstring pointing to this.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: See above comment

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: I've added a `_converter_signature()` method

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: As above

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: Code changed by refactor

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: Changed to lower case

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
markquintontulloch: Changed to creation time and `combine_json` updated to ensure earliest timestamp used from files being combined

src/python/tests/features/test_convert_to_genomio_json_base.py
markquintontulloch: Kept as is due to `FeatureConverter` being abstract

src/python/tests/features/test_convert_to_genomio_json_base.py
markquintontulloch: As above

markquintontulloch:
@JAlvarezJarreta all comments should have been addressed/answered

docs/user_guide/converter_modules.md
```diff
@@ -0,0 +1,84 @@
+# Adding a feature converter
+
+Feature converters transform a tool-specific feature file into a GenomIO JSON document. The shared command-line interface, validation, registration, and document assembly code lives in `ensembl.io.genomio.features.convert_to_genomio_json.base`; each supported tool has its own module in `src/python/ensembl/io/genomio/features/convert_to_genomio_json/`.
+
+## Implement a concrete converter
+
+Create a module named for the tool, for example `my_tool.py`. Define one `FeatureConverter` subclass for each analysis that can produce a document. A concrete converter must provide:
```
JAlvarezJarreta: I may add the link to the class's definition (if useful) under `FeatureConverter`. Or maybe an example of an existing one?

docs/user_guide/converter_modules.md
```diff
@@ -0,0 +1,84 @@
...
+    """Convert MyTool output to GenomIO JSON."""
+
+    analysis_logic_name = "my_tool"
+    analysis_display_label = "MyTool features"
+    analysis_description = "Features reported by MyTool."
+    command = "my-tool"
+    program = "my-tool"
+
+    @classmethod
+    def add_parser(cls, subparsers: argparse._SubParsersAction) -> None:
```
JAlvarezJarreta, line 41:
```suggestion
    def add_parser(cls, subparsers: argparse._SubParsersAction) -> None:
        """..."""
```
I would add something like that to indicate we expect a docstring but that this is an example. Same for the classmethod below.

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
JAlvarezJarreta: I quite like this approach ☺️

src/python/ensembl/io/genomio/features/convert_to_genomio_json/base.py
```diff
@@ -263,25 +269,65 @@ def validate_parsed_coordinates(
...
+
     if not converter.analysis_logic_name:
         raise ValueError(f"Converter {converter.__name__} has no analysis logic name")
-    registered_converter = CONVERTERS_BY_LOGIC_NAME.get(converter.analysis_logic_name)
-    if registered_converter is not None and registered_converter is not converter:
+    registered_converter = converters_by_logic_name.get(converter.analysis_logic_name)
+
+    if registered_converter is not None and _converter_signature(
+        registered_converter
+    ) != _converter_signature(converter):
```
JAlvarezJarreta, line 325: This way of `black` to divide the line annoys me more than it should.

JAlvarezJarreta (commented):
Great addition of `docs/user_guide/converter_modules.md`! And it looks good to me. Just a one minor suggestion. Happy to approve once `mypy` and `pytest` are happy.

JAlvarezJarreta (approved):
We got there at the end 🎉  Happy for you merge when you are ready.

PR #505, opened 2026-07-15
Fix FASTA splitting logic and update tests

markquintontulloch:
I found some issues with the FASTA splitting logic, addressed here.  

1. **Chunking no longer assumes the first chunk starts in an empty file.** The original implementation always split oversized sequences into chunks of max_seq_length_per_file, regardless of how much space remained in the current output file. The revised implementation calculates the first chunk based on the remaining capacity of the current file.
2. **Available space in the current output file is now utilised.** Previously, if an oversized sequence followed other sequences in a partially filled file, the remaining space was discarded because the first chunk was always written to a new file. The revised logic fills the remaining space where it is large enough to hold a meaningful chunk.
3. **Sequences that fit into a new file are no longer unnecessarily chunked.** The original implementation chunked any sequence larger than `max_seq_length_per_file`, even when the only reason it did not fit was because the current file was partially full. The revised implementation first opens a new file when appropriate and writes the sequence intact if it fits within an empty file.
4. **Chunking now correctly handles full (or overfull) output files.**  The revised implementation detects when there is no usable remaining space in the current file (including after a previous merged chunk has exceeded the configured maximum) and starts a new file before calculating chunk boundaries.
5. **The first chunk is no longer always written to a new file.**  The original implementation opened a new file whenever `writer.record_count > 0`, causing every chunk, including the first, to start in a new file. The revised implementation only opens a new file before writing subsequent chunks, allowing the first chunk to occupy any available space in the current file.
6. **Chunk merging can now collapse back to a single record.** When merging a short final chunk results in only a single chunk remaining, the revised implementation writes the original record unchanged rather than generating a single chunk record with a modified identifier. This avoids unnecessary chunking metadata when the sequence is effectively no longer split.
7. **Warning messages are now accurate.** The original implementation could incorrectly warn that a record exceeded `max_seq_length_per_file` and that chunking was disabled when the record actually fitted within the configured maximum and chunking was enabled, but the current output file lacked sufficient remaining space. The revised logic distinguishes these cases and only emits warnings when they accurately reflect the reason for the chosen behaviour.


src/python/tests/fasta/test_split.py
```diff
@@ -387,36 +388,154 @@ def test_write_record(
...
+
+    split._write_whole_record(writer, record)
+
+    writer.close()
+
+    assert filecmp.cmp(
+        out_dir / "1" / "input.1.fa",
+        data_dir / expected_fasta,
+        shallow=False,
+    )
```
JAlvarezJarreta, line 444: FYI, `ensembl.utils.plugin` has a fixture names `fixture_assert_files` that "Returns a function that asserts if two text files are equal, or prints their differences". The main difference is that the function returned does not only tell you if the files are different, but where.

src/python/tests/fasta/test_split.py
```diff
@@ -387,36 +388,154 @@ def test_write_record(
...
     out_dir = tmp_path / "out"
     out_dir.mkdir(exist_ok=True)
-    split.split_fasta(in_fasta, out_dir, **extra_args)
+    caplog.clear()
+    with caplog.at_level(logging.WARNING):
+        split.split_fasta(in_fasta, out_dir, **extra_args)
+    if expected_warning is None:
+        assert not caplog.records
+    else:
+        assert expected_warning in caplog.text
```
JAlvarezJarreta, line 538: I don't think we were testing if logs were produced before, but it is nice the machinery is there to do it easily.

JAlvarezJarreta (approved):
Looks great, cheers! 👍

PR #535, opened 2026-09-03
add community suffix for Fraxinus excelsior

Dishalodha:


JAlvarezJarreta (approved):
Don't forget to generate the new FTP YAML file and share it with Natalie so it can be updated on the FTP before this genome is released.

PR #519, opened 2026-08-03
Add literature-based genome-assembly metadata extraction module

soominhildelee:
## Summary

Adds a new `ensembl.io.genomio.literature` subpackage that extracts genome-assembly
metadata (species, ploidy, chromosome number, cultivar/strain, sex) from an
assembly's primary literature, given an NCBI assembly accession.

This work was done as part of Google Summer of Code 2026 (EMBL-EBI / Ensembl).
Mentors: Jorge Alvarez, Disha Lodha.

## What it does

Given an accession, the pipeline:
- retrieves the assembly (NCBI Datasets) and its publication (linked PMIDs,
  Entrez elink, BioProject, Europe PMC name / full-text-species / Entrez search)
  with relevance scoring;
- parses full-text XML, abstracts, and supplementary files (docx/xlsx/csv/html)
  into weighted sections;
- extracts ploidy as the INSDC 4-tuple `[level, mechanism, irregular, derivation]`
  plus chromosome number, via a rule-based extractor and a BM25 + PubMedBERT
  vector-search ensemble;
- optionally runs a grounded local Gemma-3 LLM layer (off by default) that also
  contributes sex and strain/cultivar.

Every value carries an evidence quote, method and confidence, and the pipeline
returns `null` rather than guessing when the text has no evidence.

## Added

- `src/python/ensembl/io/genomio/literature/` (fetch, parse, extract, search,
  gemma, pipeline)
- `literature_metadata` console entry point
- `literature` optional-dependencies group in `pyproject.toml`

## Current state

- Curated benchmark: 11/11 assemblies pass with the full pipeline.
- Scale test: ran over 1,059 arbitrary plant assemblies with zero unhandled
  crashes; species extracted for 98%, paper-grounded ploidy for 67%.

## Notes / to do

- Unit tests under the repo's pytest conventions are still to be added.
- Opening for review and feedback on placement and conventions.
- GSoC work-product report and dev history:
  https://github.com/soominhildelee/gsoc-genome-metadata

PR #536, opened 2026-09-07
Post-GSoC refinements: strict paper acceptance (+suggested/GoaT fallback), whole-paper Gemma, status fix, circular-import fix

soominhildelee:
Follow-up to #519, from the mentor feedback:

- Strict acceptance: only accession/BioProject-linked papers give a confirmed
  (found_in_paper) result; when none qualify, the best candidate paper is used as
  a `suggested` result, falling back to GoaT only if there's no candidate.
- GoaT cross-check flag (agree / disagree / no_reference) on every assembly.
- Gemma now reads the whole trusted paper (raised MAX_CHARS) with an accession-level
  check that it reads the correct paper.
- Result status set in run_pipeline (single-accession / Nextflow): success only when
  a ploidy level is recovered.
- Fix circular import: lazy __init__ + pipeline no longer imports the package namespace,
  so importing the package doesn't require the optional vector-search deps.

Validation on the four test genomes (rule+vector, Gemma off): 4/4 — two confirmed
(GCF_002575655.3, GCA_032464435.1), two suggested (GCA_947310255.1 -> 6, GCA_905183035.1 -> 4).

### DIFF
--- a/src/python/ensembl/io/genomio/literature/__init__.py
+++ b/src/python/ensembl/io/genomio/literature/__init__.py
@@ -21,4 +21,15 @@ number, cultivar/strain and sex via a rule + vector + optional-LLM ensemble.
 
 __all__ = ["run_batch", "run_pipeline"]
 
-from ensembl.io.genomio.literature.pipeline import run_batch, run_pipeline
+
+def __getattr__(name: str):
+    # Import from pipeline lazily (PEP 562) so that `import
+    # ensembl.io.genomio.literature` does not eagerly pull in the vector-search
+    # stack (rank-bm25 / sentence-transformers, which are optional `literature`
+    # extras) and does not create an __init__ <-> pipeline import cycle. The
+    # convenience `from ensembl.io.genomio.literature import run_pipeline` still works.
+    if name in __all__:
+        from ensembl.io.genomio.literature import pipeline
+
+        return getattr(pipeline, name)
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/src/python/ensembl/io/genomio/literature/fetch.py
+++ b/src/python/ensembl/io/genomio/literature/fetch.py
@@ -278,6 +278,27 @@ def _strongest_score(candidates: list, identifiers: dict) -> float:
     return max((score_paper_candidate(p, identifiers) for p in candidates), default=0.0)
 
 
+def _paper_is_accession_linked(paper: dict, accession: str) -> bool:
+    """Strict acceptance gate: a paper is trusted only if it is tied to THIS assembly
+    by its accession or BioProject. Papers retrieved via a direct accession/BioProject
+    link (NCBI-established) qualify on provenance; papers found by name / full-text
+    search qualify only if the assembly accession (version-less) actually appears in
+    the paper's title, abstract or full text."""
+    if paper.get("retrieval_source") in ("linked_pmid", "elink", "bioproject") or paper.get(
+        "is_reference_paper"
+    ):
+        return True
+    acc_base = accession.split(".")[0].lower()
+    text = " ".join(
+        [
+            paper.get("title", "") or "",
+            paper.get("abstractText", "") or "",
+            (paper.get("text_data", {}) or {}).get("text") or "",
+        ]
+    ).lower()
+    return acc_base in text
+
+
 def _gather_paper_candidates(accession: str, assembly: dict, identifiers: dict, max_results: int) -> list:
     """Run the Mode-B retrieval escalation and return the pooled candidate papers.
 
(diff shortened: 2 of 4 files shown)


src/python/ensembl/io/genomio/literature/fetch.py
```diff
@@ -278,6 +278,27 @@ def _strongest_score(candidates: list, identifiers: dict) -> float:
...
+    """Strict acceptance gate: a paper is trusted only if it is tied to THIS assembly
+    by its accession or BioProject. Papers retrieved via a direct accession/BioProject
+    link (NCBI-established) qualify on provenance; papers found by name / full-text
+    search qualify only if the assembly accession (version-less) actually appears in
+    the paper's title, abstract or full text."""
+    if paper.get("retrieval_source") in ("linked_pmid", "elink", "bioproject") or paper.get(
+        "is_reference_paper"
+    ):
+        return True
+    acc_base = accession.split(".")[0].lower()
```
Dishalodha, line 291: Maybe we should record the accession thats found in the paper if its different from the one provided.

src/python/ensembl/io/genomio/literature/fetch.py
```diff
@@ -278,6 +278,27 @@ def _strongest_score(candidates: list, identifiers: dict) -> float:
...
+    text = " ".join(
+        [
+            paper.get("title", "") or "",
+            paper.get("abstractText", "") or "",
+            (paper.get("text_data", {}) or {}).get("text") or "",
```
Dishalodha, line 296: are we also looking into the full text? the information can be scattered.

src/python/ensembl/io/genomio/literature/__init__.py
```diff
@@ -21,4 +21,15 @@
 
 __all__ = ["run_batch", "run_pipeline"]
 
-from ensembl.io.genomio.literature.pipeline import run_batch, run_pipeline
+
+def __getattr__(name: str):
```
Dishalodha, line 25: This raises a question for me if we are using these imports as optional? as everytime we run the cli the entry point will either run `run_pipeline` or `run_batch` which will install these imports. 

I might be wrong, feel free to comment.

src/python/ensembl/io/genomio/literature/pipeline.py
```diff
@@ -112,6 +111,21 @@ def _ploidy_recovered(result: dict) -> bool:
...
+    that is genuinely tied to THIS assembly — linked by accession (directly linked
+    PMID / Entrez elink) or by its BioProject, or with the assembly's scientific
+    name actually present in the paper text. This stops Gemma from confidently
+    reading ploidy out of an off-target paper."""
+    if paper.get("retrieval_source") in ("linked_pmid", "elink", "bioproject") or paper.get(
+        "is_reference_paper"
+    ):
+        return True
+    scientific_name = (assembly.get("scientific_name") or "").lower()
+    paper_text = (paper.get("text_data", {}).get("text") or "").lower()
```
Dishalodha, line 125: Just scientific name or maybe if any other metadata is present could make it stricter.

src/python/ensembl/io/genomio/literature/pipeline.py
```diff
@@ -145,12 +171,34 @@ def run_pipeline(accession: str, max_papers: int = 5) -> dict:
...
+        goat_level = assembly.get("reference_ploidy")
+        if goat_level is not None:
+            logger.info(f"  No qualifying paper — using GoaT reference ploidy ({goat_level}).")
+            ploidy = {
+                "level": goat_level,
+                "mechanism": None,
+                "status": "from_reference",
+                "source": "GoaT",
+                "method": "reference_taxon",
+                "confidence": 0.6,
```
Dishalodha, line 192: Sorry why have we used 0.6 as confidence here?

src/python/ensembl/io/genomio/literature/pipeline.py
```diff
@@ -213,9 +267,14 @@ def run_pipeline(accession: str, max_papers: int = 5) -> dict:
     #     retrieved vector chunks, so it forms an independent opinion even when the
     #     decisive ploidy sentence didn't make the vector-search cut. combine_with_gemma
     #     folds its result in as a third vote. No-op unless GEMMA_ENABLED=1.
-    if gemma.is_enabled():
+    if is_enabled() and not _is_correct_paper_for_accession(best_paper, assembly):
+        logger.info(
+            "  [gemma] skipped: the selected paper is not confidently tied to this "
+            "accession (no accession/BioProject link and scientific name absent from text)"
+        )
+    elif is_enabled():
```
Dishalodha, line 275: would be useful to make it a cli parameter to enable it

src/python/ensembl/io/genomio/literature/fetch.py
soominhildelee: Good idea, done. When the paper cites an assembly accession whose base differs from the one we queried (a GCF/GCA twin or a different version), I now stamp it onto the paper as `other_accessions_in_text` so a curator can see we matched on a related-but-different id.

src/python/ensembl/io/genomio/literature/fetch.py
soominhildelee: Yes: the match string joins `title` + `abstractText` + `text_data.text` (the full text when we have it), then lowercases, so a scattered accession mention anywhere in the body still counts.

src/python/ensembl/io/genomio/literature/__init__.py
soominhildelee: You're right that the CLI entry point always ends up importing the vector stack, so this isn't making those deps optional at CLI runtime. The lazy `__getattr__` is only there to (a) break the `__init__` ↔ `pipeline` import cycle and (b) let `import ensembl.io.genomio.literature` (or importing a single sibling submodule, e.g. in tests/tooling) succeed without eagerly pulling in `rank-bm25`/`sentence-transformers`. Running the pipeline still requires them, as expected.

src/python/ensembl/io/genomio/literature/pipeline.py
soominhildelee: Agreed. I added the strongest signal first: if the assembly accession itself appears in the paper text we trust it; the scientific-name match is now the weaker fallback. Happy to fold in strain/cultivar too if you'd like it stricter still.

src/python/ensembl/io/genomio/literature/pipeline.py
soominhildelee: 0.6 is chosen to sit on the pipeline's existing confidence scale rather than being arbitrary. Paper-derived ploidy (rule + vector, and Gemma) is capped up near 0.95–0.98 when the level is actually stated in this assembly's own paper, and even a suggested answer from a candidate paper is scaled by 0.7. A GoaT value is weaker than all of those: it is a curated species-level reference, not evidence read from this assembly's paper, and it is only used when no qualifying paper exists at all. So 0.6 places it clearly below any paper-backed result while keeping it well above a null/"no evidence" outcome, which lets a curator sort or threshold on confidence and immediately tell reference-inferred values apart from literature-confirmed ones. I've pulled it into a named constant `GOAT_REFERENCE_CONFIDENCE` with this rationale as a comment; happy to tune the exact number if you'd prefer a different cut-off.

src/python/ensembl/io/genomio/literature/pipeline.py
soominhildelee: Added a `--gemma` flag that flips `GEMMA_ENABLED=1,` so it reuses the same switch `is_enabled()` already reads without threading a parameter through `run_pipeline`/`run_batch`.

PR #538, opened 2026-09-17
ENSGENOMIO-26 Red output to JSON conversion

markquintontulloch:


src/python/tests/features/test_convert_to_genomio_json_red.py
```diff
@@ -0,0 +1,142 @@
...
+        {
+            "seq_region": "chr2",
+            "seq_region_start": 31,
+            "seq_region_end": 31,
+            "seq_region_strand": "+",
+            "repeat_start": 1,
+            "repeat_end": 1,
+            "repeat_consensus": expected_consensus_key,
+        },
+    ]
```
JAlvarezJarreta, line 99: Just my two cents: whilst this works, consider using `DeepDiff` in some of these cases for 2 reasons:
* It will add to the error message what was different (easier to spot the issue)
* If the dictionary starts holding objects as values the test will keep working, whilst I think `==` will not

src/python/tests/features/test_convert_to_genomio_json_red.py
```diff
@@ -0,0 +1,142 @@
...
+
+    with pytest.raises(
+        ValueError,
+        match=r"^Found \d+ errors while parsing Red output in .*:",
+    ) as excinfo:
+        red.parse_output(rpt_path)
+
+    error_message = str(excinfo.value)
+    for expected_fragment in expected_fragments:
+        assert expected_fragment in error_message
```
JAlvarezJarreta: Just a thought: why not build the full regex `match` with the expected content rather than checking that `ValueError` was raised first, then extracting the error message and checking again parts of it?

JAlvarezJarreta (approved):
Looks good to me, thanks. Just a couple of comments, but approving in case you want to disregard them (they are not critical).

vsitnik (approved):
ok. thank you!

JAlvarezJarreta (approved):

